Issue #5, opened 2019-07-03
Develop assertion model for screen reader testing


mcking65:
Objective: Determine how gramular to make assertions when testing screen reader support for an element or widget pattern.

mcking65:
Proposing a less gramular model where:

1. There are two kinds of tests: Reading and Operation.
2. Each test may be performed with one of two pre-conditions: 1) SR in reading mode or 2) SR in Interaction mode
3. Each test is associated with an element or element type, e.g., checkbox, group, menubar, menuitem.
4. Each test has an importance (Must Have | Should Have | Nice to Have)
5. There is one assertion for a given test type, pre-condition, element, and importance. This assertion may cover multiple ARIA attributes. The test may be executed in multiple ways.

Attaching a spreadsheet with with examples. So far, only includes checkbox testing.
[TestSimplificationExploration.xlsx](https://github.com/w3c/aria-at/files/3354188/TestSimplificationExploration.xlsx)

mcking65:
Started testing menubar and experimented with some other ideas on presentation. See attached workbook with new tab for navigation menubar.
[TestSimplificationExploration.xlsx](https://github.com/w3c/aria-at/files/3378375/TestSimplificationExploration.xlsx)

mcking65:
Made more progress refining naming tests and defining assertions. Also made more friendly column (less technical) column names in spread sheet for menubar. Added more columns for notes about results of tests. See menubar tab in attached workbook.
[TestSimplificationExploration.xlsx](https://github.com/w3c/aria-at/files/3427219/TestSimplificationExploration.xlsx)

mfairchild365:
This is looking good @mcking65 - some comments:

* I like the names of the columns and agree that they are more friendly
* I like the addition of the "Success Notes" column. That should help future testers identify past behavior and compare their new results. It also helps whoever is reading the report understand how the screen reader behaves.
* It might be good to consider renaming the "tested attributes" column to something like "tested attributes and elements". I'm thinking of a case where the ARIA example uses an HTML element for its implicit roles and attributes. If I remember correctly, the [simple two-state checkbox example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-1/checkbox-1.html) included a `<ul>` element.
[...]

PR #12, opened 2019-10-23
Initial checkbox tests

spectranaut:
This is a first draft of what we discussed, "WPT-like tests". The following has been committed:
* 3 test files in `tests/checkbox/` that cover the 6 tests that @mcking65 wrote for the two state checkbox example ([linked in this comment](https://github.com/w3c/aria-at/issues/5#issuecomment-514649962)). 
* 4 files in the `tests/checkbox/reference` sub-directory that contain two HTML pages (and a shared css and javascript file) that the AT is tested against in the three test files. In order to help the test stand alone, and reduce long term confusing dependencies, I simply copied APG's example code for the "two state checkbox" into a "reference" directory. When writing the tests, I broke the example up into two HTML pages (one HTML page has one checkbox, the other is an exact copy of the two-state checkbox example) in order to make the assertions easier to write. EDIT: this has been changed, we are using the APG examples directly.

I'll call the WPT-like test file (with steps and assertions) the "test file", and the html page that it references the "test page". Multiple test files might reference a single test page.

The test files are written using an imaginary helper file called `resources/ariaatharness.js`. The imaginary harness file has the following API:
* `openTestPage`: opens a specified test page in a new window.
* `executeScriptInTestPage`: If you need to get the test page into a specific state before a test, then you have the option to script it instead of asking the tester to change the state.
* `presentATTest`: this function abstracts away the presentation of the instructions, assertions and recording of results. It takes the following arguments:
   * `mode`: a list of AT modes that this test should be preformed under. Currently we have discussed `reading` and `interaction` modes. I'm not convinced this should be a list, but for checkbox, the reading and interactive mode tests are the same.
   * `action`: this is the "abstract operating instruction" we have discussed, that can be used to describe an action that a user can take with any specific AT.
   * `assertions`: a list of human understandable assertions about what should happen after the action.
   * `onPartialFailure`: the purpose of this function is to help with branching. If there is a failure, the test author can specify what questions are asked after the failure.

Most of the work is in `presentATTest`. I imagine it does the following:
1. Gets and displays a set of specific operating instructions ("commands") for an AT from the supplied "mode" and "action". The specific AT will be provided by the test runner.
2. Displays the "assertions" to the tester.
3. Displays a form with "pass" or "fail" for the tester.
4. If "fail" is selected, then we can have branching. At this point, we want the tester to be able to mark which assertions and which commands fail. Initially, I imagined you would see a list of "commands", and you could mark a "command" as failing. If the test author wanted more details about the failure, they could supply a "onPartialFailure" function that will supply more questions about a command's failures if a command is marked as failing (for example, which assertion did the command fail).

Some questions I have: 
1. How do we want tor record failures?
2. How should we display tests if the test can be run in both modes? Should we instead have different test files (that are identical except for the "mode" argument)? Should we otherwise say if a test can be preformed in both or one mode, and have the test runner supply that information?
3. Should the assertions be structured objects instead of strings. More test writing will illuminate this, I think.
4. "Importance" (must have/should have) not encoded in tests.


tests/checkbox/read-checkbox.html
```diff
@@ -0,0 +1,70 @@
...
+<link rel="help" href="https://w3c.github.io/aria/#checkbox">
+<link rel="help" href="https://w3c.github.io/aria/#aria-checked">
+<script src="/resources/ariaatharness.js">
+<script>
+
+  openTestPage("reference/two-state-checkbox.html");
+
+  presentATTest({
+    mode: ["reading", "interaction"],
+    action: "navigation",
```
spectranaut: We should reconsider "action" as a name.

tests/checkbox/read-checkbox-grouping.html
```diff
@@ -0,0 +1,37 @@
...
+         <label for="role-${failedInstruction}">Role was not announced.</label>
+       </div>
+       <div>
+         <input type="checkbox" id="name-${failedInstruction}" name="name-${failedInstruction}" value="Name was not announced.">
+         <label for="name-${failedInstruction}">Name was not announced.</label>
+       </div>
+       <div>
+         <input type="checkbox" id="boundaries-${failedInstruction}" name="state-${boundariesInstruction}" value="Boundaries were not announced.">
+         <label for="state-${failedInstruction}">Boundaries were not communicated.</label>
+       </div>
```
mfairchild365: Consider tracking what the output was that resulted in a failure.

For this, you might need:
 1. the exact command used
 2. the exact output (or summary of output)
 3. Other relevant information. For example, did this happen when reading forwards through the document or backwards? For boundaries, what part of the boundary was not conveyed (enter or exit), etc.

tests/checkbox/read-checkbox-grouping.html
```diff
@@ -0,0 +1,37 @@
...
+  openTestPage("reference/checkbox-group.html");
+
+  presentATTest({
+    mode: ["reading", "interaction"],
+    action: "navigation",
+    assertion: [
+      "The role 'group' is announced",
+      "The group's name 'Sandwich Condiments' is announced",
+      "The boundaries of the group (before the first checkbox and after the last checkbox) are communicated"
+    ],
```
mfairchild365: I like this approach since it is simple and intuitive, and some good suggestions were provided on the call earlier. Some of my thoughts:

1. This could end up being a rather large list of commands. To save time, it might be worthwhile to select 2 or 3 common commands and only test those for the early phases of this project. I think that would have the highest return on investment.
2. When a command fails, how will we track that specific failure (see next comment).
3. Abstracting to other assistive technologies (beyond screen readers): I took the approach of keeping the assertions generic and applying them to many technologies. For example, both screen readers and voice control software make use of a control's name, but voice control software will not convey boundaries. With this approach, some assertions might not be applicable to all AT, but it does let us display support for a specific aspect of the feature across many AT.
4. if one of the end goals is to display support data for different aria attributes, we should consider how assertions for specific attributes will be reused across tests. I'd suggest that the assertions should be defined in a centralized place, and tests should link to them as foreign keys.

tests/checkbox/read-checkbox-grouping.html
spectranaut: So in this iteration, I imagine that presentATTest initially provides "passed" or "failed" below the instructions of the test. If the tester clicks 'failed', then the presentATTest provides the list of commands (perhaps as checkboxes) for the tester to select the exact command that failed. That commands is the "failedInstruction" that is passed as an argument to this function. Selecting a command as failing would then reveals these extra questions. I think I should rename the "failedInstruction" as "failedCommand", but it does assume just one command (rather than a series of commands, I don't know if our tests will have series of commands). Does this cover your concern for exact command used?

That is a really good point about boundaries, I'll just added it to this test. As for exact output I wonder if that is something we would want for every failed command, so we don't need to supply it in this function (because it can be hard coded into the logic of presentATTest).

tests/checkbox/read-checkbox-grouping.html
spectranaut: Good point about assertions. I was thinking the assertions should be more structured, at least. We might want to decide how to abstract them/write them/link to them after we write some more tests. Can you link some examples of your more generic assertions?

tests/checkbox/read-checkbox-grouping.html
spectranaut: So in this iteration, I imagine that presentATTest initially provides "passed" or "failed" below the instructions of the test. If the tester clicks 'failed', then the presentATTest provides the list of commands (perhaps as checkboxes) for the tester to select the exact command that failed. That commands is the "failedInstruction" that is passed as an argument to this function. Selecting a command as failing would then reveals these extra questions. I think I should rename the "failedInstruction" as "failedCommand", but it does assume just one command (rather than a series of commands, I don't know if our tests will have series of commands). Does this cover your concern for exact command used?

That is a really good point about boundaries, I'll just added it to this test. As for exact output I wonder if that is something we would want for every failed command, so we don't need to supply it in this function (because it can be hard coded into the logic of presentATTest).

tests/checkbox/read-checkbox.html
spectranaut: used "task" for now and used a verb/noun phrase instead of only a noun.

tests/checkbox/read-checkbox-grouping.html
mfairchild365: Yes, I think that makes sense. One question that I have... if multiple commands fail, but each yield different output, is it beneficial to record the output for each failed command?

spectranaut: There might be no way of knowing until we've written enough tests and run though them -- but this is a variable we should remember to play around with! Seems like each command could fail differently, though.

tests/checkbox/read-checkbox-grouping.html
mfairchild365: Sure. Using a11yspport.io and HTML links as an example:

- [an HTML link feature that references many tests](https://a11ysupport.io/tech/html/a(href)_element)
- [The assertions are defined on the feature object (JSON)](https://github.com/accessibilitysupported/a11ysupport.io/blob/master/data/tech/html/a(href)_element.json)
- [And referenced by the tests (JSON)](https://github.com/accessibilitysupported/a11ysupport.io/blob/master/data/tests/tech/html/links/example1.json)
- Note that abstract versions of common assertions such as 'convey name', and 'convey role' [are defined further left in the build process](https://github.com/accessibilitysupported/a11ysupport.io/blob/master/src/feature-helper.js#L94). This helps with consistency and reduces redundancies. 
- My approach is slightly different than the approach we are taking here, but I think we can apply similar concepts. We will be testing the same ARIA features across many tests.

mfairchild365:
@spectranaut I'm not seeing some of the changes that you showed us in this PR. has it been updated?

spectranaut:
@mfairchild365 I'm going to merge this PR into a feature branch and then open a new one with my changes since, to preserve the conversation history!

Issue #14, opened 2019-10-27
Mapping abstract test instructions to concrete, screen-reader-specific instructions


jfhector:
This first issue serves as a summary of the thread.

##### Testing commands for checkbox

* [JAWS checkbox testing commands](https://nov13-testing-commands.now.sh/checkbox/JAWS)
* [NVDA checkbox testing commands](https://nov13-testing-commands.now.sh/checkbox/NVDA)
* [VoiceOver checkbox testing commands](https://nov13-testing-commands.now.sh/checkbox/VoiceOver)

##### Testing commands for menubar

* [JAWS menubar testing commands](https://nov13-testing-commands.now.sh/menubar/JAWS)
* [NVDA menubar testing commands](https://nov13-testing-commands.now.sh/menubar/NVDA)
* [VoiceOver menubar testing commands](https://nov13-testing-commands.now.sh/menubar/VoiceOver)

#### Outstanding decisions

* We will need to decide which testing commands to include in version 1, and which ones to leave out to avoid making the tests to effortful

#### Next steps

* I believe that the testing commands for both checkbox and menubar are ready to use for the prototype. It'd be good for someone else to review them at some point in the near future and raise issues.

jfhector:
I've taken a stab at creating a document that maps abstract test instructions to screen-reader-specific concrete instructions, starting with the Checkbox example.

Here is the document in Excel format:
[Checkbox testing commands.xlsx](https://github.com/w3c/aria-at/files/3775922/Checkbox.testing.commands.xlsx)

#### What the document does

The document gives a list of concrete, screen-reader-specific commands for testers to execute for each test for the Checkbox example.

So, for example, for "Read checkbox", it shows what concrete commands to press for JAWS, NVDA and VoiceOver in the different interaction modes.

#### Structure of the document

The document has three tabs:

1. Checkbox testing commands with JAWS
2. Checkbox testing commands with NVDA
3. Checkbox testing commands with VoiceOver macOS

Each tab contains one table that gives specific testing commands for that particular screen reader, for each screen reader mode, for each Checkbox test. 

#### Where this data comes from

I've put together these commands from:
* [Matt's 'Test Simplification Exploration' spreadsheet from early July](https://github.com/w3c/aria-at/issues/5#issuecomment-510122453)
* Deque's [Screen Reader Reference Guides](https://dequeuniversity.com/screenreaders/)

To avoid mistakes, I've also tested these commands in JAWS, NVDA and VoiceOver today.

[...]

PR #17, opened 2019-11-05
Proof of concept test harness

spectranaut:
This test harness is a proof of concept, so the code is not final. In writing the harness I had to change the API from the previous tests in #12 .

Some questions:
- [ ] What to do about modes -- should we include different modes in the same test? For example, first run the test in reading mode, then run the test in interactive mode, every time? 
- [ ] What information do we want to save on failure and on success?

You can look at this version of test harness here:
- [read-checkbox.html](https://spectranaut.github.io/aria-at/tests/checkbox/read-checkbox.html)
- [operate-checkbox.html](https://spectranaut.github.io/aria-at/tests/checkbox/operate-checkbox.html)
- [read-checkbox-grouping.html](https://spectranaut.github.io/aria-at/tests/checkbox/read-checkbox-grouping.html)

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -1,0 +1,209 @@
+import {isKnownAT, getATCommands, getModeInstructions} from './at-commands.mjs';
+
+const DEFAULT_AT = 'jaws';
+const DEFAULT_RESULTS = ['Success', 'Complete Failure', 'Partial Failure'];
+
+const TEST_HTML_OUTLINE = `
+<section id='errors' style='display:none'><h2>Test cannot be performed due to loading error(s).</h2></section>
+<section id='instructions'></section>
+<section id='record-results'>
+</section>
+<section id='submit-results'>
+</section>
+`;
+
+let handleResult;
+
+export function openTestPage(page) {
+  console.log(`TODO: open ${page}`);
+}
+
+export function executeScriptInTestPage(page) {
+  console.log(`TODO: execute script in page.`);
+}
+
+export function presentATTest(test) {
+  let title = document.title;
+  document.open();
+  document.write(TEST_HTML_OUTLINE);
+  document.close();
+  document.title = title;
+
+  let at = DEFAULT_AT;
+  const mode = test.mode[0];
+  const modeInstructions = getModeInstructions(mode, at);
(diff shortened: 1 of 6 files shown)


tests/resources/at-commands.mjs
```diff
@@ -0,0 +1,89 @@
...
+  jaws: 'JAWS'
+}
+
+/* Add new commands here */
+
+const AT_COMMAND_MAP = {
+  "navigate to checkbox": {
+    reading: {
+      jaws: [
+        "Insert+Tab",
```
mfairchild365: It might be useful to include both the command name and the keys to execute the command. As written, I could see it being a bit confusing to testers, especially if they have not used a screen reader before? "Okay, I see that I'm supposed to press Insert+Tab, but what does that DO? And how is "up/down" different?

Also, up and down arrow keys MAY yield different results. So it might be good to break them into their own commands. Same with tab/shift tab and left/right.

tests/resources/aria-at-harness.mjs
```diff
@@ -0,0 +1,372 @@
...
+  <td>
+    <input type="checkbox" id="pass-${c}-${a}" class="pass" name="pass-${c}-${a}">
+    <label for="pass-${c}-${a}">Pass</label>
+  </td>
+  <td>
+    <input type="checkbox" id="fail-${c}-${a}" class="fail" name="fail-${c}-${a}">
+    <label for="fail-${c}-${a}">Fail</label>
+  </td>
+  <td>
+    <input type="text" id="utterance-${c}-${a}">
```
mfairchild365: As mentioned on the call, there are a few a11y issues here. Consider running the page through axe to help identify, fix, and validate. https://www.deque.com/axe/

tests/resources/at-commands.mjs
spectranaut: good points, @mfairchild365, I'm going to fix these things in another issue after we settle on which commands to test in #14

tests/resources/aria-at-harness.mjs
spectranaut: Great tip thanks axe was really helpful! I fixed all issues except for the accessible name for the radios.. I want to wait for Matt's feedback on what the accessible names should be.

Issue #36, opened 2020-01-22
Review and test the checkbox example for the prototype


mfairchild365:

mfairchild365:
@mcking65 - I started to review the checkbox tests and have the following notes.

1. The exact path of execution is not always clear. For example, the test "Reading checked checkbox in interaction mode" has two instructions. The first is to put the screen reader into forms mode. The second is to read the checkbox with the supplied commands. This makes sense, however, in order to create consistent, easily understandable, and reproducible results, I think we need an additional step that answers the question 'how did the user navigate to the checkbox'?. In other words, it's missing an instruction that says something like "press the tab key until the first checkbox is focused".  Exactly how the user navigated to an element may impact how the element was read. I've seen this in previous versions of JAWS. This is also true for the operation tests. Note that there are situations where keyboard focus may remain on a previous input, while virtual focus is moved to a new input, which can cause some screen readers to convey incorrect information. I believe this was a problem in JAWS 2019 and was fixed in the 2020 version. All of that to say: providing detailed instructions on the path of execution to get to the desired element is pretty important.
2. Reading mode commands are listed for "Navigating to checkbox in interaction mode." Actually, "Navigating to checkbox in reading mode" describes interaction mode instructions, so I think the names just got swapped.

mfairchild365:
I'll make a PR to add missing navigation instructions

mfairchild365:
Also, virtual is misspelled

mfairchild365:
also change "spoken" to "conveyed"

mfairchild365:
Update: I submitted two pull requests related to the issues outlined here.

However, I also noticed that the number of steps is [hardcoded in the test harness](https://github.com/w3c/aria-at/blob/master/tests/resources/aria-at-harness.mjs#L184).

I think it would still be good to add an instruction like "click the button to open the test page and wait for the first checkbox to automatically receive focus", but I'm not sure about the best way to do that yet given the architecture.

spectranaut:
I think this review has been completed!

PR #39, opened 2020-01-30
fixes related to the checkbox tests

mfairchild365:
See #36


spectranaut:
Oh, whoops -- @mfairchild365, I didn't notice the "WIP"

mfairchild365:
@spectranaut - this should be good to merge now. There may be some more changes, but we can make a new PR for those if needed.

PR #40, opened 2020-01-30
fix executeScriptInTestPage

mfairchild365:
JS injection into examples was failing, which was causing check boxes to not receive focus when the test page was opened. This was happening in Chrome; I didn't check other browsers.

After debugging, I discovered that (at least in chrome) the new window first loads the `about:blank` page and then loads the proper URL. Therefore when we check the `readyState`, it was evaluated as `complete`, but the window was still pointed to `about:blank`.  So our code was being injected to the `about:blank` page, and was obviously failing.

 This simply adds another check to ensure that the window's `origin` matches the same `origin` as the parent window before executing our script in the test page.

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -113,7 +113,9 @@ function putTestPageWindowIntoCorrectState() {
 function executeScriptInTestPage() {
   let setupTestPage = allBehaviors[currentTestedBehavior].setupTestPage;
   if (setupTestPage) {
-    if (testPageWindow.document.readyState !== 'complete') {
+    if (testPageWindow.location.origin !== window.location.origin // make sure the origin is the same, and prevent this from firing on an 'about' page
+        || testPageWindow.document.readyState !== 'complete'
+    ) {
       window.setTimeout(() => {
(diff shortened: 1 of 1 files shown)


zcorpan:
> This simply adds another check to ensure that the window's `origin` matches the same `origin` as the parent window before executing our script in the test page.

I was going to say that this won't make any difference, because `about:blank` pages have the same origin as their creator. But the code does `location.origin`, which returns the origin of the *URL*, which works since the `about:blank` URL's origin is `null`. Assuming that the test's URL's origin is not also `null`, which it can be if it's loaded from a `file:` URL (at least in Firefox). It's [undefined what the origin of a `file:` URL is](https://url.spec.whatwg.org/#origin).

Checking if `location.href` is `about:blank` might be more reliable.

mfairchild365:
Thanks! That's a good point and sounds like it would be more reliable. I had assumed that using file URLs to access the project is not supported for the prototype, but it sounds like that is not the case.

zcorpan:
I don't know if we want `file:` to be supported. 🙂In fact it probably already doesn't, since module scripts require CORS headers, which `file:` can't have.

So we're good here, I think!

Issue #6, opened 2019-07-03
Design a simple way for testers to contribute


jfhector:
Here are some thoughts and [a small design exploration](https://test-page-jfariaatjuly3.now.sh/) from me. This is based on the [assertion model that @mcking65 shared in a separate issue](https://github.com/w3c/aria-at/issues/5), and the conversation that we had reviewing it.

#### I do agree that we need to make things simpler and easier to understand for testers

@mfairchild365 This week I ran the [test for `aria-details` on a11ysupport.io](https://a11ysupport.io/tests/tech__aria__aria-details/run) to familiarise myself with the data model and interface.

I struggled to figure out exactly how to perform the test. In particular:
  * After reading the assertions, I wasn't sure exactly how I should perform the test, and what constituted success or failure.
  * I was confused when prompted to select what command I used to perform the test, because I expected to be told what commands I should use. I wasn't sure which option was the right one for me to select.
  * I didn't easily understand the options available to me when inputting the position of the virtual cursor before and after the command (eg. 'target', 'start of target', ...). I expected to be told, and felt concerned that I might not perform the right test in the right way.

From that perspective, I think that the conversation we're having now about how to simplify things for users is very helpful and needed.

I also realise that the test pages and instructions on a11ysupport.io are generated efficently at scale, and that clearer and more granular instructions might not be viable.

#### My thoughts on the simplifications that @mcking65 has been exploring

Here's the link again to the [assertion model that @mcking65 shared in a separate issue](https://github.com/w3c/aria-at/issues/5). My comments here focus not so much on the data model itself, but on how testing instructions can be presented.

###### 1. I like that Matt's table gives more specific instructions about test methods.

I think that it's important to tell testers what they need to do and how, rather than asking them what they did.

###### 2. I like the idea of describing expected results in non-technical terms as much as possible.

I think that assertions should be written to help people who are not fully confident in their knowledge of accessibility.

###### 3. I like the idea of organising testing instructions and results in a table.

Seeing the different tests in different rows, with instructions on the left and results on the right, felt very intuitive to me. It instantly gave me an clear mental model of how to use the interface (ie. the table).

###### 4. I like the idea of wrapping different 'test methods' into one single test.

I like how a tester is expected to test how a checkbox is announced, when reached in different ways, in one go. I imagine that following this idea will help us simplify the interface, without making things harder for users. (Caveat: I don't fully yet fully understand the downsides of such a simplification for our data).

##### What I think we can simplify further

Here are a few suggestions:

[...]

PR #49, opened 2020-02-10
Menubar edit

jongund:
Added initial tests adn at-commands for the menubar-editor example


jongund:
@mcking65 @spectranaut @mfairchild365 

Can you take a look at this pull request, it uses two scripts in `/tests/menubar-edit/data` directory to generate the test files and adds commands to the `/test/resources/at-commands.mjs` file.  The data comes from 3 CSV files that were exported from an excel spreadsheet.

jongund:
Fixed some bugs, the command `npm run review-tests` is now at least generating review files.

tests/resources/at-commands.mjs
```diff
@@ -1,259 +1,662 @@
...
+        keys.INSERT_UP,
+        keys.INSERT_TAB,
+        keys.INSERT_UP,
+        keys.INSERT_TAB,
+        keys.INSERT_UP
+      ]
+    },
+    interaction: {
+      jaws: [
+        keys.INSERT_TAB,
```
spectranaut: somehow this script is adding things multiple times! gotta remove these all and fix the script.

tests/menubar-editor/reading-text-color-group-reading.html
```diff
@@ -0,0 +1,32 @@
...
+
+<script type="module">
+  import { verifyATBehavior, displayTestPageAndInstructions } from "../resources/aria-at-harness.mjs";
+
+  verifyATBehavior({
+    mode: ["reading"],
+    task: "Reading Text Color group",
+    specific_user_instruction: "read the Text Color group",
+    output_assertions: [
+      [1, "when focus is on a menuitem in Text Color group(Black); read the group(Text Color)"],
```
spectranaut: This seems more like an instruction than an assertion. It's possible this should be the "specific_user_instruction"

tests/menubar-editor/data/menubar-edit-tests.csv
```diff
@@ -0,0 +1,28 @@
+Status,Test No,Title,Mode,Testing Task,specific_user_instruction(Task),output_assertion(Success Criteria),2,3,4,5,6
+Asking clarification for a group(1/30),1,Activating 'Text Formatting' menubar switches mode from reading to interaction,reading,activate menubar,activate the 'Text Formatting' menubar,the screen reader switches to mode that interaction,,,,,
+Asking clarification for a group(1/30),2,Navigating to 'Text Formatting' menubar switches mode from reading to interaction,reading,navigate to menubar with keys that switch modes,navigate to the 'Text Formatting' menubar,After command; JAWS is in Forms Mode,,,,,
+,3,Navigating to menubar,reading,navigate to menubar ,navigate to the first menuitem(Font) in menubar,The role 'menubar' is conveyed,The name 'Text Formatting' is spoken,The role 'menuitem' is conveyed,The name 'Font' is spoken,The state of the checkbox aria-haspopup is conveyed,The position in menubar and number of menubar items is spoken
```
mfairchild365: I haven't fully reviewed these CSV files, but I did notice "The state of the checkbox aria-haspopup is conveyed". The attribute `aria-haspopup` allows for many values. I don't think it is clear what exactly is expected here. IMO, to pass, the screen reader must convey that the menuitem opens a menu. As written, a tester might interpret this to mean any value associated with `aria-haspopup` may pass.

Additionally, I'm not sure what "checkbox" is referring to.

tests/menubar-editor/data/create-at-commands.py
```diff
@@ -0,0 +1,169 @@
+import sys
```
mfairchild365: I don't mind mixing languages (NodeJS + Python) for fast prototyping, however, I'd like to avoid doing this in production and sick with one language. The more dependencies the higher the barrier to contribution.

tests/menubar-editor/navigating-to-menubar-interaction.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
```
mfairchild365: If these files are being generated from a csv, it might be better to ignore these files in git and generate them in the build process. That would help to reinforce the csv as the authoritative source and prevent someone from modifying the generated files.

tests/resources/at-commands.mjs
spectranaut: oh, I missed this was fixed in the last commit!

tests/menubar-editor/navigating-to-menubar-interaction.html
spectranaut: I agree this is fine for now, but we should revisit the whole test creation workflow. Hopefully after we get this first set of tests are in we can do some strategizing here (informed by this first set of test writing), and Bocoup can spend some time on it in the next month.

We have one tricky area -- right now, most of the data is declarative, but there is the optional `setupTestPage` script to get the widget in the correct state before the test. I don't think we want to write javascript in a spreadsheet or CSV file. It's possible we could use the .html test files as an input, along with the csv files, as well as an output, for a test generation script. For example: If an html file exist with that name, we can rewrite all the declarative pieces of information from a CVS file, and keep the `setupTestPage` from the test file itsself. Otherwise, create a new html page.

tests/menubar-editor/navigating-to-menubar-interaction.html
```diff
@@ -0,0 +1,36 @@
...
+      [1, "The role 'menubar' is conveyed"],
+      [1, "The name 'Text Formatting' is spoken"],
+      [1, "The role 'menuitem' is conveyed"],
+      [1, "The name 'Font' is spoken"],
+      [1, "The state of the checkbox aria-haspopup is conveyed"],
+      [1, "The position in menubar and number of menubar items is spoken"]
+    ]
+  });
+
+  displayTestPageAndInstructions("reference\menubar-edit.html");
```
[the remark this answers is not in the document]
spectranaut: These need to be regenerated like this:
```
```

spectranaut:
In the last two commits I did the following to `menubar-editor/data/create-at-commands.py`:
1. I used "os.path" to create the relative file path so that I can run the script on linux :)
2. I used the `specific_user_instruction(Task)` (column #6) for the specific_user_instruction, and excluded it from the assertion list. I think it was just an off-by-one error in both cases.

There is one thing that still needs to be addressed before we merge this PR:
- [ ]  The command script is adding multiple rows, or maybe it adds rows every time it is run. I thought it  would be quicker if @jongund fixes this (see my comment here: https://github.com/w3c/aria-at/pull/49#discussion_r377388373)

jongund:
@spectranaut 

I was going to work on some changes based on the feedback so far and regenerate the files.
The names will change to do edits to the spreadsheet.

Jon

tests/menubar-editor/navigate-to-menuitem-in-menubar-using-interaction-mode.html
```diff
@@ -0,0 +1,31 @@
...
+<title>Navigate to menu item in menubar using interaction mode</title>
+<link rel="author" title="Valerie Young" href="mailto:valerie@bocoup.com">
+<link rel="help" href="https://w3c.github.io/aria-practices/examples/menubar/menubar-2/menubar-2.html">
+<link rel="help" href="https://w3c.github.io/aria/#menuitem">
+
+<script type="module">
+  import { verifyATBehavior, displayTestPageAndInstructions } from "../resources/aria-at-harness.mjs";
+
+  verifyATBehavior({
+    setupTestPage: function setupTestPage(testPageDocument) {
```
spectranaut: We should update "testPageDocument" to "document" in order for this script to work.

tests/menubar-editor/data/at-commands.mjs
```diff
@@ -0,0 +1,434 @@
...
+        keys.SPACE
+      ]
+    }
+  },
+  "change state of menuitemcheckbox in submenu using reading mode": {
+    reading: {
+      jaws: [
+        keys.ENTER,
+        keys.SPACE,
+        keys.ENTER,
```
spectranaut: This is still duplicated multiple times!

tests/menubar-editor/change-state-of-menuitemcheckbox-in-submenu-using-reading-mode.html
```diff
@@ -0,0 +1,26 @@
...
+    mode: ["reading"],
+    task: "change state of menuitemcheckbox in submenu using reading mode",
+    specific_user_instruction: "uncheck the 'Italic' menu item in the 'Style/Color' submenu",
+    output_assertions: [
+      [1, "The unchecked state of the checkbox is conveyed"],
+      [1, "Screen reader switching to interaction mode is conveyed"]
+    ]
+  });
+
+  displayTestPageAndInstructions("reference\menubar-editor.html");
```
spectranaut: This blackslash needs to be updated to a forward slash, because this file is opened in a browser.

tests/menubar-editor/data/create-test-files.py
```diff
@@ -0,0 +1,124 @@
...
+  return a
+
+def getSetupTestPageScript(fname):
+
+  code = ''
+
+  if len(fname.strip()):
+    f = open(os.path.join('js', (fname + '.js')), 'r')
+
+    code = 'setupTestPage: function setupTestPage(testPageDocument) {\n'
```
spectranaut: this needs to be changed from `testPageDocument` to be `document` in order to be consistent with the scripts.

tests/menubar-editor/navigate-to-menubar-in-reading-mode.html
```diff
@@ -0,0 +1,29 @@
...
+<title>Navigation to menubar using reading mode</title>
+<link rel="author" title="Valerie Young" href="mailto:valerie@bocoup.com">
+<link rel="help" href="https://w3c.github.io/aria-practices/examples/menubar/menubar-2/menubar-2.html">
+<link rel="help" href="https://w3c.github.io/aria/#menubar">
+
+<script type="module">
+  import { verifyATBehavior, displayTestPageAndInstructions } from "../resources/aria-at-harness.mjs";
+
+  verifyATBehavior({
+    setupTestPage: function setupTestPage(testPageDocument) {
```
spectranaut: As you can see here, the scripts expect "document", so we should replace "testPageDocument" with document.

tests/menubar-editor/data/test.template
```diff
@@ -0,0 +1,22 @@
...
+<meta charset="utf-8">
+<title>%TITLE%</title>
+<link rel="author" title="Valerie Young" href="mailto:valerie@bocoup.com">
+%REFERENCES%
+<script type="module">
+  import { verifyATBehavior, displayTestPageAndInstructions } from "../resources/aria-at-harness.mjs";
+
+  verifyATBehavior({
+    %SETUP_TEST_PAGE%
+    applies_to: ["%APPLIES_TO%"],
```
spectranaut: In a different PR we are about to merge (https://github.com/w3c/aria-at/pull/61), we are change this list to a single string, the mode will be "reading" or "interaction". I'll merge that PR this afternoon, can we remove the brackets here?

spectranaut:
Menubar update: should use "conveyed" not "spoken" in all assertions.

PR #60, opened 2020-02-11
Add 'applies_to' key to keep track of which AT the test tests

spectranaut:
Hey @jongund -- I added an `applies_to`, like we discussed in the email. You could just edit your template and add a new row to include this information. For almost all the tests we write the value will be `["Desktop Screen Readers"]` unless the test only applies to a subset of screen readers, in which case it should list the screen readers it applies too, such as: `["NVDA", "JAWS"]`


jongund:
@spectranaut 
It looks good to me, I will apply to my template and test files

PR #70, opened 2020-02-19
Checkbox updated 2

mfairchild365:


tests/checkbox/operate-checkbox-interaction.html
```diff
@@ -14,8 +14,7 @@
     task: "operate checkbox",
     specific_user_instruction: "check and uncheck the first checkbox",
     output_assertions: [
-      [1, "The change in the state of the checkbox is conveyed when the box becomes unchecked."],
-      [1, "The change in the state of the checkbox is conveyed when the box becomes checked."]
+      [1, "Change in state is conveyed"]
```
spectranaut: I know we talked about this in the CG meeting, but was this for sure the conclusion? What if the state is only not convey when you go one direction? Or is that so unlikely, we shouldn't worry about it?

mfairchild365: I share the same concern, however it would be rare enough that a tester could describe which value failed in the notes.

spectranaut: No need to bike shed so lets just go with this and change the test if it is insufficient :)

PR #71, opened 2020-02-20
Menubar edit script and data fixes

spectranaut:
Hi @jongund , I did two minor things in this PR: 

1. The key "mode" is now a string instead of a list, I simply updated the template to get this change.
2. The "reference" entry of the "reference" CSV file and tab in the excel sheet had a backslash instead of a forward slash. I updated that in both places as well.

I'll merge this later today so that when you work on the changes in language tomorrow, if you have an chance to take a look at it that be great!

Right now, we are having a conflict where my edits on a linux computer ending lines with "lf" and your changes on a windows computer ending lines with "crlf". I'm not sure the best way to handle these conflicts - have you had to handle this in the past?


jongund:
@spectranaut 

I will look at your changes, there are GitHub configuration options to handle the line endings:
https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings

jongund:
@spectranaut 

What branch should I be working in with my edits?

spectranaut:
@jongund are you working right now? you can make a new branch from master, but probably only after I merge this PR into master. Or, if you want, you can push to this branch.

jongund:
@spectranaut 

I will wait an make a new branch after you merge

Jon

PR #72, opened 2020-02-21
updated test files for menubar editor

jongund:
@mcking65 
@spectranaut 

1. I updated the test files for your review.
1. I am optionally generating a "index.html" file with links to the test files, it only works with localhost, I don't know if it will mess up the testing harness by having the file in that directory.  If so just delete it.  It is a little funky since it does not work with the test server, only local host, but you cn look at and see if it might be useful for us.  It was useful to me to get quick references to individual test files.
1. I am out of town for the next week, so I can make changes when I get back, Yohta maybe able to update the CSV files used to generate the test
1. The startup scripts don'e seem to be working, so I will need to talk to Valerie when I get back to learn what I am doing wrong


spectranaut:
If you want to easily see and test the test files in the browser, @jongund, you can just run `python -m SimpleHTTPServer` in the test directory, and you should be able to navigate to [http://localhost:8000/menubar-editor/](http://localhost:8000/menubar-editor/) in the browser and see a directory listing! I'm going to delete the index.html file, because it will be treated as a test file because it ends in ".html" by the runner.

PR #73, opened 2020-02-22
Refactor test design to allow commands.json file in each pattern directory

spectranaut:
From now on, we will simply add a `commands.json` file with the tasks/mode/assistive technology to command list data structure.


tests/resources/at-commands.mjs
```diff
@@ -1,443 +1,69 @@
 import * as keys from './keys.mjs';
 
-const KNOWN_ATS = {
-  jaws: 'JAWS',
-  nvda: 'NVDA',
-  voiceover: 'VoiceOver'
-}
+export class commandsAPI {
```
evmiguel: Is the `commandsAPI` the pattern we are going to use in the production system?

tests/menubar-editor/commands.json
```diff
@@ -0,0 +1 @@
...bar using interaction mode": {"interaction": {"jaws": [["TAB_AND_SHIFT_TAB"]]}}, "read menuitemradio in a group in submenu using reading mode": {"reading": {"jaws": [["INS_TAB"], ["INS_UP"]]}}, "read menuitemradio in a group in a submenu using interaction mode": {"interaction": {"jaws": [["INS_TAB"], ["INS_UP"]]}}, "read menuitemcheckbox in a submenu using interaction mode": {"interaction": {"jaws": [["INS_TAB"], ["INS_UP"]]}}, "change state of menuitemcheckbox in submenu using reading mode": {"reading": {"jaws": [["ENTER"], ["SPACE"]]}}, "read menu item in menubar using reading mode": {"reading": {"jaws": [["INS_TAB"], ["INS_UP"]]}}, "navigate to menuitemradio in submenu using interaction mode": {"interaction": {"jaws": [["ENTER"], ["SPACE"], ["DOWN"]]}}, "navigate to menuitem in menubar using reading mode": {"reading": {"jaws": [["LEFT_AND_RIGHT", " (with Smart Navigation on)"]]}}, "change state of menuitemradio in submenu using reading mode": {"reading": {"jaws": [["ENTER"], ["SPACE"]]}}, "navigate to menuitemradio in submenu using reading mode": {"reading": {"jaws": [["UP_AND_DOWN"]]}}, "open submenu of menubar using reading mode": {"reading": {"jaws": [["ENTER"], ["SPACE"]]}}}
```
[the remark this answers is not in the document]
spectranaut: It could, but I did have the suggestion/thought from Mike that if it is printed like this then people might realize it is generated and not for manual editing.

tests/resources/at-commands.mjs
spectranaut: Right now, this file is part of the test suite, it's for rendering the test files. So only indirectly, because the production system consumes the tests

tests/menubar-editor/commands.json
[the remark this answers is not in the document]
evmiguel: It wasn't clear to me. Is there a way for us to note this difference?

tests/resources/at-commands.mjs
evmiguel: In this case, can we use a more descriptive class name or have a JSDoc about what this class does? `commandsAPI` wasn't clear enough for me to understand until I saw it in use.

spectranaut: good point. There should be documentation. I'll add it right now!

tests/menubar-editor/commands.json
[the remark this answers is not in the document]
spectranaut: very soon all of these files will be generated, so there won't be a need for the distinction!

evmiguel:
After discussion with @spectranaut, in the future, the `commands.json` and test files will be generated. The official process of how these files get generated has yet to be solidified, but will be decided in the near future. For now, both generated and manually created `commands.json` will be committed to this repo.

tests/resources/at-commands.mjs
```diff
@@ -1,443 +1,103 @@
...
-  }
-  return '';
-}
-
-export function isKnownAT(at) {
-  return KNOWN_ATS[at.toLowerCase()];
-}
+import * as keys from './keys.mjs';
+
+/** Class for getting AT-specific instructions for a test against a design pattern. */
```
spectranaut: I added comments @evmiguel thanks for that nudge

PR #92, opened 2020-02-28
Add missing keys

spectranaut:
@mfairchild365 we need to add some keys constants for the tests you modified, just as a heads up!


mfairchild365:
Thanks! I actually did that but just forgot to commit it. Doh!

spectranaut:
lol whoops :)

PR #128, opened 2020-03-19
Update menubar setup scripts

jongund:
ready for review.

Is it OK to add an object "menubarEditor" to the "document" object for use in the setup scripts.


zcorpan:
Can you clarify which commits have changes that don't come from aria-practices? Or which files need a closer review?

> Is it OK to add an object "menubarEditor" to the "document" object for use in the setup scripts.

I don't think there's a policy for this yet for this repo, but I think it's better to have a global object and not extend documents or elements.

jongund:
@zcorpan 

There does not seem any other way to get access to the test object, since only the document object is passed in.  So maybe there should be a second parameter with a reference to the testing object.

zcorpan:
You can get to the global with `document.defaultView`. But yeah, explicitly passing in the information needed in parameters, or an object as a single parameter, seems better.

@spectranaut thoughts?

jongund:
@zcorpan Maybe put this on the agenda for Wednesday, since it will be important for creating the ARIA practices examples.

zcorpan:
@jongund this has merge conflicts now. Can you resolve them?

jongund:
@zcorpan The merge conflicts have been fixed, should be ready to go.

jongund:
@zcorpan Any luck trying to get this merged?

zcorpan:
There was a merge conflict with `tests/menubar-editor/reference/js/menubar-2-init.js`. As far as I can tell it should be removed, so I resolved the conflict by removing it.

Issue #108, opened 2020-03-06
Decide on list of "applied_to" options


spectranaut:
We have been using both lists of specific screen reader and also categories.

- Specific ATs: 'JAWS', 'NVDA', 'VoiceOver', and 'Orca'
- Categories: 'Screen readers', 'Desktop screen readers'

Do we want to have a mix of categories and specific lists? If so, should we decide some norms for the relationships between the categories and specific screen readers?

mcking65:
Yes, there are specific relationships; it is a hierarchy of assistive technologies. This will become more important as we add more assistive technologies.

Here are the values and hierarchy I propose:

* Screen Readers
   * Desktop Screen Readers
      * JAWS
      * NVDA
      * VoiceOver for MacOS
      * Narrator
      * Orca
      * ChromeVOX
   * Mobile Screen Readers
      * VoiceOver for iOS
      * TalkBack
      * VoiceAssistant

I prefer that we write "VoiceOver for MacOS" rather than MacOS VoiceOver" because it is usually more convenient to hear the screen reader name first.

When validating and using the applies to field, if the field contains "Screen Readers", then the assertions apply to all screen readers. If it contains "Desktop Screen Readers", then they apply to all desktop screen readers. We need to be able to provide comma separated value lists as well, such as "JAWS, NVDA"".

Values like "JAWS, NVDA, Mobile Screen Readers" could also be possible but are not likely.

spectranaut:
To do for Val: integrate this information into the test design.

PR #144, opened 2020-04-01
Update support file with 'applies_to' information

spectranaut:
Adding more information to this file that I think would be necessary in order to consume and run these tests by any third party! The extra information is the relationship between "applies_to" values in the tests.

What values can be put in the "applies_to" field was discussed in this issue: https://github.com/w3c/aria-at/issues/108

But I do have one question. We see in the tests an applies_to value of "VoiceOver". Is there a chance, @mcking65, that we should have a category call "VoiceOver" the refers to both "VoiceOver for MacOS" and "VoiceOver for iOS" -- or should those tests be update to revert to just "VoiceOver for MacOS"? These are "Test 2" and "Test 4" in the [combobox test plan](https://w3c.github.io/aria-at/review/combobox-autocomplete-both.html).

### DIFF
--- a/tests/support.json
+++ b/tests/support.json
@@ -1,7 +1,7 @@
 {
   "ats": [
     {
-      "name": "VoiceOver for MacOS",
+      "name": "VoiceOver for macOS",
       "key": "voiceover"
     },
     {
@@ -15,12 +15,12 @@
   ],
   "applies_to": {
     "Desktop Screen Readers": [
-      "VoiceOver for MacOS",
+      "VoiceOver for macOS",
       "NVDA",
       "JAWS"
     ],
     "Screen Readers": [
-      "VoiceOver for MacOS",
+      "VoiceOver for macOS",
       "NVDA",
       "JAWS"
     ]
--- a/tests/support.json
+++ b/tests/support.json
@@ -2,7 +2,7 @@
   "ats": [
     {
       "name": "VoiceOver for macOS",
-      "key": "voiceover"
+      "key": "voiceover-macos"
     },
     {
       "name": "NVDA",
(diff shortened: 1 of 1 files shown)


tests/support.json
```diff
@@ -13,13 +13,28 @@
...
+    ],
+    "Screen Readers": [
+      "VoiceOver for MacOS",
+      "NVDA",
+      "JAWS"
+    ],
+    "VoiceOver": [
+      "VoiceOver for MacOS"
+    ]
+  ],
```
zcorpan, line 30:
```suggestion
  "applies_to": {
    "Desktop Screen Readers": [
      "VoiceOver for MacOS",
      "NVDA",
      "JAWS"
    ],
    "Screen Readers": [
      "VoiceOver for MacOS",
      "NVDA",
      "JAWS"
    ],
    "VoiceOver": [
      "VoiceOver for MacOS"
    ]
  },
```

zcorpan:
@mcking65 says we don't need VoiceOver as a category.

tests/support.json
```diff
@@ -13,13 +13,25 @@
       "key": "jaws"
     }
   ],
+  "applies_to": {
+    "Desktop Screen Readers": [
+      "VoiceOver for MacOS",
```
mcking65, line 18: I've learned the official way to right this is with a lowercase 'm' -- macOS

tests/support.json
```diff
@@ -13,13 +13,25 @@
...
+      "NVDA",
+      "JAWS"
+    ],
+    "Screen Readers": [
+      "VoiceOver for MacOS",
```
mcking65, line 23: ditto

zcorpan:
I've fixed spelling of macOS and changed the key to `voiceover-macos`

zcorpan:
@spectranaut I assigned to you to merge

Issue #145, opened 2020-04-01
Test Refactor


spectranaut:
In consuming the tests, I've realized we need to do a refactor of the test format of the test `.html` files. The purpose of the refactor is to more easily consume metadata from the tests. The refactor is ultimately pretty small, but it will effect three things in this repository. The changes that will have to be made:

- Jon's csv files to test files script ([scripts/create-tests.js](https://github.com/w3c/aria-at/blob/master/scripts/create-tests.js))
    - This change will be very straight forward, because the tests files have a template. The information is the same, but the template will have to change.
    - Changing this script will allow us to update all the test files to the new format in one go, which is very exciting!
- The aria-at-harness.mjs file ([tests/resources/aria-at-harness.mjs](https://github.com/w3c/aria-at/blob/master/tests/resources/aria-at-harness.mjs))
    - The logic is almost entirely the same, but the entry point to the logic will have to change. Instead of getting that information from the `VerifyATBehavior` call, the script will look for a `<script id ="test-data" type="application/json"></script>` and initiate the test from there.
    - The javascript `setupTestPage` code will have to be escaped to to include it in the JSON. Since these test files will most likely never be written by hand, I think that is reasonable. We will then have to inject it into the popup window with the APG example in a different way, maybe appending the popup html with a  `script` tag with the stringified javascript inside of it.
- The script that makes the "review test plan" pages ([scripts/test-reviewer.mjs](https://github.com/w3c/aria-at/blob/master/scripts/test-reviewer.mjs)).
    - I think this script should be removed. Instead, I think Jon's script should make "review test plan" pages at the same time that it produces the test files.
    - If we don't remove this script, then we need to update it to consume the new test format instead of the old test format. Should be straight forward, as the change makes the data more structured and easy to parse.

And example of the new format I am imaging for the test files is provided at the end of this issue. For comparison, see the original format for this test in: [read-checked-checkbox.html](https://github.com/w3c/aria-at/blob/7bd2e959ba9cd3dc11b81a480dae5089f84aa883/tests/checkbox/read-checked-checkbox.html).

The change moves all the test specific information from a javascript script tag (where all the test specific information is passed to the harness through an API call) to a json script tag. When designing these test, it seemed like we might want to support a lot more logic in the test file, but it turns out the only javascript you need is pretty minimal. Moving all the declarative parts of the test to the JSON script tag I think simplifies the test file and makes it very straightforward to consume in the runner.

```
<!DOCTYPE html>
<meta charset="utf-8">
[...]

PR #149, opened 2020-04-03
Use JSON to pass test data to the harness

mfairchild365:
See #145 

- [x] Get feedback on this approach
- [x] Move JSON to an external file
- [ ] Use a single HTML file for all tests and load test-dependant data via the approach outlined in #145? might move this to another PR depending on complexity.
- [x] Make sure that the script to generate the review files works correctly

@spectranaut @zcorpan I'd like your feedback on this approach. I'm generating a `scripts.js` file that defines a single object with a property for each 'script', and then I tell the harness which property to use instead of encoding the JS itself in the JSON data.

### DIFF
--- a/scripts/create-tests.js
+++ b/scripts/create-tests.js
@@ -341,7 +341,9 @@ function createTestFile (test, refs, commands) {
     id = '0' + id;
   }
   let testFileName = 'test-' + id + '-' +cleanTask(test.task).replace(/\s+/g, '-') + '-' + test.mode.trim().toLowerCase() + '.html';
+  let testJSONFileName = 'test-' + id + '-' +cleanTask(test.task).replace(/\s+/g, '-') + '-' + test.mode.trim().toLowerCase() + '.json';
   let testFileAbsolute = path.join(testDirectory, testFileName);
+  let testJSONFileAbsolute = path.join(testDirectory, testJSONFileName);
 
   if (typeof test.setupScript === 'string') {
     let setupScript = test.setupScript.trim();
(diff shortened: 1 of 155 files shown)


zcorpan:
I like that. 👍

mfairchild365:
Update

- Updated to move JSON out of HTML files, and fixed the test-review script
- I think we will need to tackle the task to use a single HTML file in another PR
- Some test-specific data still exists in the HTML file, such as `<link>` references.

zcorpan:
> I think we will need to tackle the task to use a single HTML file in another PR

That sounds OK to me.

scripts/create-tests.js
```diff
@@ -367,20 +369,24 @@ function createTestFile (test, refs, commands) {
...
   import { verifyATBehavior, displayTestPageAndInstructions } from "../resources/aria-at-harness.mjs";
 
-  verifyATBehavior();
+  fetch("${testJSONFileName}")
+    .then(response => response.json()) // parse the JSON from the server
+    .then(data => {
+      verifyATBehavior(data);
+    });
+  
   displayTestPageAndInstructions("${refs.reference}");
```
zcorpan, line 390: Should this happen after `verifyATBehavior` has been called?

zcorpan:
I tried to run this

```
$ node scripts/create-tests.js tests/checkbox-gen/
```

and then

```
npm run start
```

Then opening http://localhost:3000/aria-at/ and trying to review or run the new test, I get

> Cannot GET /aria-at/tests/checkbox-gen/test-01-navigate-into-checkbox-group-interaction.html

Am I holding it wrong? :-)

mfairchild365:
@spectranaut 

1. Can you review this?
2. Are the 404 issues related to project configuration or something that I changed? If its project configuration, what do I need to change?

zcorpan:
I discussed with @spectranaut now and the 404 issue is a known bug when running the prototype locally. To test, you need to change static.config.js as follows

```
diff --git a/static.config.js b/static.config.js
index 6447f21..0506344 100644
--- a/static.config.js
+++ b/static.config.js
@@ -113,6 +113,6 @@ export default {
     require.resolve('react-static-plugin-sitemap'),
   ],
   siteRoot: "https://w3c.github.io",
-  basePath: "aria-at",
-  devBasePath: "aria-at",
+  basePath: "",
+  devBasePath: "",
 }
```
and then shut down the server and `npm run start` again. The links will be broken, so need to manually remove `aria-at/` from the URLs, but with that, it works. In any case, this should work (without changing static.config.js) on GitHub Pages since it has its root at `/aria-at/`.

This PR should be rebased on master and then check in the new generated files. I can do this.

mfairchild365:
Thank you for the config info and thank you for generating the test files as part of this PR, @zcorpan.

PR #153, opened 2020-04-08
Update menubar setup scripts

jongund:
I fixed the problem with index.html, 

The other file was not in my branch, so not sure why there is a conflict it should already be deleted.


zcorpan:
Can you rebase on current master? #128 was merged but it looks like this PR has some of the same commits.

tests/menubar-editor/index.html
```diff
@@ -48,6 +48,7 @@
   <h1>Index of Test Files</h1>
   <p>This is useful for viewing the local files on a local web server and provides links that will work when the local version of the
   test runner is being executed, using <code>npm run start</code> from the root director: <code>C:\inetpub\wwwroot\aria-at\</code>.</p>
+<<<<<<< HEAD
```
zcorpan, line 51: This should be removed, right?

zcorpan:
So this was like #128 but is in a confused state. Let's close.

jongund:
@zcorpan It will be fixed when we get all the new CVS files into master and then can regenerate all the test and index files, so I would just not worry about it right now.

I have anew branch called 'update-all-tests' that I will submit a PR once the new `update-checkbox` branch gets merged to master.   Then will generate all new test files and index files with the latest stable creation script.

PR #154, opened 2020-04-08
Correct breadcrumb link to ARIA-AT Home page from the Results page

Malvoz:
Currently points to https://w3c.github.io/ as opposed to https://w3c.github.io/aria-at/.

### DIFF
--- a/src/pages/results.js
+++ b/src/pages/results.js
@@ -22,7 +22,7 @@ export default function ResultsPage() {
         <title>ARIA-AT: All Test Results</title>
       </Head>
       <nav aria-label="Breadcrumb">
-        <a href="../">ARIA-AT Home</a>
+        <a href="/aria-at/">ARIA-AT Home</a>
       </nav>
       <h1>Test Run Results</h1>
       <ul>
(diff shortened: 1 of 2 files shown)


zcorpan:
So in https://w3c.github.io/aria-at/results/ it links to `../../`

in https://w3c.github.io/aria-at/results/test-run/1 it also links to `../../`

The correct URL would be `/aria-at/`, or `../` for the first and `../../../` for the second.

Also, https://w3c.github.io/aria-at/results/ links to https://w3c.github.io/aria-at/results/test-run/1 , but reloading that page or navigating back and forth in history results in https://w3c.github.io/aria-at/results/test-run/1/ (with a trailing slash), causing relative URLs to change their destination.

Properly fixing this might be a bit of effort, but this is the *prototype* and will soon be replaced by a new system. I wonder if we can remove the breadcrumbs altogether?

src/pages/results.js
```diff
@@ -22,7 +22,7 @@ export default function ResultsPage() {
         <title>ARIA-AT: All Test Results</title>
       </Head>
       <nav aria-label="Breadcrumb">
-        <a href="../../">ARIA-AT Home</a>
+        <a href="../">ARIA-AT Home</a>
```
zcorpan, line 25:
```suggestion
        <a href="/aria-at/">ARIA-AT Home</a>
```

zcorpan:
OK I've fixed both breadcrumbs. The trailing slash issue is still there, though.

PR #170, opened 2020-05-01
Checkbox: Test navigating through group with arrows separately from other commands plus some other fixes

mcking65:
Structural changes:
1. Added tests for navigating through the checkbox group with arrow keys (navigate sequentially) that are separate from commands that skip over boundaries (e.g., tab). This enables required assertions about boundaries for sequential navigation.
2. Experiment: separated VoiceOver into separate tests so that we can remove mode from the title.
3. Changed test sequence so all the simple checkbox tests are first then the grouping related commands are at the end.

Other Changes:
1. Fixed assertion wording where state of 'not checked' was specified on some tests where the checkbox is checked.
2. Added F and Shift_F for JAWS and NVDA commands that navigat in reading mode
3. Removed X and shift_X from NVDA commands for navigating out of a checkbox group
4. Removed CTRL_OPT_CMD_J	and SHIFT_CTRL_OPT_CMD_J from VoiceOver commands for navigating OUT of a checkbox group.
5. Removed INS_UP from NVDA and JAWS commands for read checkbox group
6. Removed JAWS and NVDA laptop key map keys (are not available by default)
7. Removed periods from the end of some titles
8. Editorial revisions to some instructions for clarity.

Finally, Re-ran create-tests.js for checkbox.


mcking65:
When I re-ran create-tests, I got the following:

> ***
>[Command]: The key reference "SHIFT_F" is invalid for the "navigate into checkbox group" task.
>[Command]: The key reference "SHIFT_F" is invalid for the "navigate into checkbox group" task.
>[Test 16]: Reference does not exist: group,aria-labelledby
>[Test 17]: Reference does not exist: group,aria-labelledby
>[Test 18]: Reference does not exist: group,aria-labelledby

I have no idea how to fix the shift_f error. I tried fixing the aria-labelledby error by adding aaria-labelledby to the references.csv, but that didn't help.

mfairchild365:
@mcking65 - I pushed a fix for the issues you were seeing when generating the tests. I also updated the generated tests.

Two things about the new tests which do not include the mode in the title:

* After building the project, I was able to run the modeless tests with JAWS and NVDA. Is that expected?
* I find that not having a mode in the title is confusing. At a glance, not having a mode in the title looks like a bug when all of the other tests include the mode. I think it might be better to name it 'modeless' or something similar. Thoughts?

mcking65:
@mfairchild365 commented:

> @mcking65 - I pushed a fix for the issues you were seeing when generating the tests. I also updated the generated tests.

Awesome, thank you!!

> Two things about the new tests which do not include the mode in the title:
> 
> * After building the project, I was able to run the modeless tests with JAWS and NVDA. Is that expected?

Oh no ... that is not expected. Why is that happening if the applies to field has only VoiceOver? Is that a bug in the harness?

> * I find that not having a mode in the title is confusing. At a glance, not having a mode in the title looks like a bug when all of the other tests include the mode. I think it might be better to name it 'modeless' or something similar. Thoughts?

Where do the test titles appear side-by-side? It would seem very strange to me to mention a mode for a screen reader that does not have a mode. That is the confusion I am hoping to eliminate.

The new production test queue does not show test titles, only the test plan run name, AT, and browser. When you look at a test, it is displayed for only one AT. I guess when we review plans for all AT one time they are consecutive in the report. However, it makes sense that the test is appropriately matched to the AT.

Thinking ahead, there are lots of AT where the concept of mode does not apply. It just happens to be fundamental in JAWS, NVDA, and Narrator.

I'd like to be able to run the tests locally so I can review the generated output. I don't know how to do that.

mcking65:
Curious if people have thoughts on the addition of the concept of sequential navigation as a separately tested set of behaviors in this plan.

mfairchild365:
> I'd like to be able to run the tests locally so I can review the generated output. I don't know how to do that.

There is a bug in the runner which makes this a bit difficult. See https://github.com/w3c/aria-at/pull/149#issuecomment-614300404

In essence, there are two lines at the end of `static.config.js` that need to be changed. Remove the string `aria-at` from both `basePath` and `devBasePath`. Then when the project loads in your browser, you will have to remove the `aria-at/` path from the URL in your browser every time you navigate to a new page.

spectranaut:
> > Two things about the new tests which do not include the mode in the title:
> > 
> > * After building the project, I was able to run the modeless tests with JAWS and NVDA. Is that expected?
> 
> Oh no ... that is not expected. Why is that happening if the applies to field has only VoiceOver? Is that a bug in the harness?

Oh, I remember this. "Applies to" concept was added on too late for me to incorporated into the "aria-at-harness.js" code. So instead it's the runners job to not display tests that don't have "applies_to" (and the prototype runner doesn't, though the production runner does! :) ). We could update the "aria-at-harness" to do something other than display the test if the test doesn't apply to the screen reader you are trying to test.

PR #181, opened 2020-05-16
Set the default to JAWS even if `at` is not defined in the URL

### LINKED ISSUE #180: index.html links does not work
I am working updating the menubar-editor tests and I did a merge from master.
When I try to check the updated tests using the pages referenced in the `index.html` file in the `menubar-editor` directory the links reference broken test pages.  It seems the error in  generating a test page is related to specifying an Assistive Technology.  Not sure where this should be in the script.

mfairchild365:
fixes #180


jongund:
@mfairchild365 
Can we just add the params to the URL: 
* `[url]?at=jaws`
* `[url]?at=nvda`
* `[url]?at=voiceover_macos`

PR #183, opened 2020-05-19
Update readme with more explanation about the project

isaacdurazo:
I'm submitting this PR on behalf of @zcorpan, who has come up with a more thorough description of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,14 @@
 # ARIA and Assistive Technologies (ARIA-AT)
 
-This repo contains a test suite and harness for assessing AT support of ARIA. These tests are designed to be modular and embeddable in higher-level applications.
+This repo contains a test suite and harness for assessing assistive technology (AT)  support of ARIA. These tests are designed to be modular and embeddable in higher-level applications. A web app for managing  test cycles and reporting test results ([ARIA-AT-App](https://github.com/bocoup/aria-at-app)) is also being developed and is planned to be available on w3.org in June 2020.
 
 ## Scope and Method
 
-The initial scope of this test suite is manual tests for screen readers. Tests are written as assertions against accessible design patterns [examples from the ARIA Practices Guide](https://w3c.github.io/aria-practices/examples/).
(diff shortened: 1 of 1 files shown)


zcorpan:
Thanks @isaacdurazo! This should be ready for review @mfairchild365 @mcking65

README.md
```diff
@@ -1,15 +1,36 @@
...
+## Documentation
 
 [Documentation is available in the wiki](https://github.com/w3c/aria-at/wiki).
+
+## Get involved
+
+You can participate in the following:
+
+* Join the [community group](https://www.w3.org/community/aria-at/)
+* [Sign up as a tester](https://github.com/w3c/aria-at/issues/162) for the Pilot Test (May 27 - June 2)
```
mfairchild365, line 22: To keep the readme from changing frequently, it might be better to link to a generic 'Sign up as a tester' wiki page, which we update every time we are about to run a new testing cycle. Thoughts?

README.md
```diff
@@ -1,15 +1,36 @@
-# Assistive Technology ARIA Experience Assessment Project
+# ARIA and Assistive Technologies (ARIA-AT)
 
-The aim of this project is to facilitate consistently effective rendering of ARIA-enabled UIs by assistive technologies by:
+This repo contains a low-level test suite and harness for assessing the interoperability of AT interpretations of ARIA. These tests are designed to be modular and embeddable in higher-level applications.
```
mfairchild365, line 3: Instead of "for assessing the interoperability of AT interpretations of ARIA",
I suggest "for assessing AT support of ARIA".

That would be a bit simpler.

zcorpan, line 3:
```suggestion
This repo contains a test suite and harness for assessing AT support of ARIA. These tests are designed to be modular and embeddable in higher-level applications.
```

README.md
zcorpan: We don't have such a wiki page, and I'd like to land this change asap. We can do that in the next iteration though!

PR #184, opened 2020-05-19
Menubar assertion updates

jongund:
@mfairchild365 
@zcorpan 
I have made a few changes to the following files:
* Updated the creation script `create_tests.js` to include links for testing each "applies to" assistive technology in the `index.html` file.  See a [preview of the updated index file](https://raw.githack.com/w3c/aria-at/menubar-assertion-updates/tests/menubar-editor/index.html).
* Updated `aria-at-harness.mjs` to parse the "Instructions" content in the CSV file to include multi-step instructions.  No modifications needed for existing CSV files, it uses a `|` character to break the instructions up into multiple list items in the test page.
* Added instructions for several menubar tests
* Working on testing all the tests


mfairchild365:
@jongund - do I understand correctly that you expect to make a few more changes? Please let me know when this is ready for a review.

jongund:
@mfairchild365 
Just need to check the VoiceOver keyboard mappings, and then it will be ready for review.
Jon

jongund:
@mfairchild365 
@zcorpan 
@mcking65 
The updated menubar editor tests are ready for review.
Test 14 for VoiceOver seems a little out of scope, but want other peoples input into this type of test.

mfairchild365:
Thank you, Jon. This is wonderful progress! I didn't have time to do a full review, but I wanted to get you some feedback as soon as possible. I focused the most on Test 14 since you specifically mentioned that.

## General observations:

1. Several of the interaction mode tests appear to apply to VoiceOver. As with the checkbox test plan, we should move these into their own test files since VoiceOver does not have an interaction mode.

## Test 1

The assertion states "Screen reader switching to interaction mode is conveyed by opening the menu item" - should this say "when opening the menu item"? An earcon usually indications the mode switch, not the fact that a menu was opened. See test two as an alternative wording for the assertion. 

## Test 14

1. The test tile includes the word 'interaction mode'. If I recall correctly, we decided not to name modes in VO tests.
2. Step two says to turn off quick nav mode. This should be off by default, and I don't think we should mention it at all. I think this is in some if not all tests.
3. Commands like `VO + Down Arrow` do not seem to be working. Did you mean `VO + Right Arrow`?
4. The assertions state "The name 'Black' is conveyed" but the instructions tell me to navigate to the "Blue" menu item. This is also an issue in test 13.
5. The assertion states that the menu item should be checked, but we navigated to an unchecked menu item. This is also an issue in test 13.
5. I think there is value in this test, but I think the instructions and title need work.

mfairchild365:
For whatever it's worth, I'm okay with merging this and revisiting some of the tests/instructions/assertions in another PR.

PR #187, opened 2020-05-20
[WIP] Link to the test page from review test plan

zcorpan:
@spectranaut I don't know how to fix this. Can you help?

I suppose it could link to the reference page (e.g. "Checkbox Example (Two State)"), rather than running the test with `at=jaws` etc. Is the reference page's path accessible from here?

### DIFF
--- a/scripts/review-template.mustache
+++ b/scripts/review-template.mustache
@@ -75,8 +75,8 @@
       <ul>
         <li>Mode: {{mode}}</i>
         <li>Applies to: {{allReleventATs}}</li>
-        <li>Lasted edited: {{lastEdited}}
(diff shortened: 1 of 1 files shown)


spectranaut:
These files are hosted at: 

`https://w3c.github.io/aria-at/review/<design-pattern>.html`

The test files are hosted at: 

`https://w3c.github.io/aria-at/tests/<design-pattern>/<test-file>?at=<at-key>`

So can you not put a relative path: `../tests/<design-pattern>/<test-file>?at=<at-key>` ?

zcorpan:
How do you get the right `<design-pattern>` part into the mustache template?

zcorpan:
Let's close this and create a wiki page instead.

zcorpan:
https://github.com/w3c/aria-at/wiki/Test-plans-that-are-ready-for-external-review

PR #192, opened 2020-05-21
Test Harness: add showResults query string and postMessage communication

s3ththompson:
This PR adds two features necessary to support bocoup/aria-at-app#103 (Editing test results and saving partial form values):
- ability to prevent the test harness from replacing the test contents html with the results table html, with a new `showResults` query string parameter (default value is `true` to maintain current behavior)
- `postMessage` communication to send two messages to any parent window of the test: a `loaded` message and a `results` message (this allows the parent app which embeds the test as an iFrame to cleanly know when async test loading has finished and when results have been submitted)

Note: the `postMessage` communication is sent to parent windows of any origin in order to support any app that may embed aria-at tests. This is safe as long as the test results do not contain user or privacy sensitive data.

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -654,8 +654,6 @@ function submitResult(event) {
     status: overallStatus
   };
 
-  appendJSONResults(data);
-
   // send message to parent if test is loaded in iFrame
   if (window.parent && window.parent.postMessage) {
     window.parent.postMessage({
@@ -669,6 +667,8 @@ function submitResult(event) {
   if (showResults) {
     showResultsTable();
   }
+
+  appendJSONResults(data);
 }
 
 
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -65,9 +65,11 @@ const PAGE_STYLES = `
 
 let behavior;
 let behaviorResults;
+let overallStatus;
 const errors = [];
 let testPageUri;
 let testPageWindow;
+let showResults = true;
 
 let at;
 let commandsData;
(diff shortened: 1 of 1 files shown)


tests/resources/aria-at-harness.mjs
```diff
@@ -584,6 +607,7 @@ function submitResult(event) {
 	// If any priority 1 assertion fails, the test fails for this command
 	if (priority === 1) {
 	  support = 'FAILING';
+    overallStatus = 'FAIL';
```
spectranaut, line 610: spacing ^

tests/resources/aria-at-harness.mjs
```diff
@@ -623,13 +648,32 @@ function submitResult(event) {
...
+
+  // send message to parent if test is loaded in iFrame
+  if (window.parent && window.parent.postMessage) {
+    window.parent.postMessage({
+      type: 'results',
+      data: data
+    }, '*');
+  }
+
   endTest();
```
spectranaut, line 667: You call end test, but reviewed the function?

tests/resources/aria-at-harness.mjs
```diff
@@ -715,15 +755,10 @@ function showUserError() {
   }
 }
 
-function reportResults(testResults, status) {
+function appendJSONResults(data) {
```
spectranaut, line 758: When testing in the prototype, results are not getting put into a script tag -- which breaks the prototype runner.  I think it this function is not being called, but haven't dug in. You can test if you want by running a server from the /tests/ directory like (`python -m SimpleHTTPServer `) without the show results = false -- then complete the test, and see that the script tage doesn't appear.

tests/resources/aria-at-harness.mjs
s3ththompson: the original `endTest` became three separate functions: `showResultsTable` (handles replacing HTML), `endTest` (handles closing test popup window), and `appendJSONResults` (handles inserting JSON data into script element). (The earlier bug was that `appendJSONResults` was called before  `showResultsTable` which replaced the documents entire html! 🤦)

PR #197, opened 2020-05-26
Menubar test: Add VoiceOver specific tests plus bug fixes

mcking65:


mcking65:
we will review in master from prod runner and then make necessary fixes. All current tests except for last one have VO equivalent. A few more tests to add that are not in plan as well to come in next PR.

PR #201, opened 2020-05-27
Harness: Add screen reader config instruction

mcking65:
Modifies instructions on each test page to include a step for restoring default configuration. Points to a wiki help page. Just a sort of hack to get us by for now.

Also makes one small grammar adjustment related to the setup scription description.


mcking65:
@spectranaut We might need to regenrate all test scripts after merging this. However, we need to get the other PRs for the test changes merged before doing that. This one should take only a few seconds to review and merge.

PR #266, opened 2020-06-04
Bump the limit of test assertions to 30

### LINKED ISSUE #193: Is there a limit of 5 assertions for each test?

mfairchild365:
fixes #193

Notes

* Bumps limit from 5 to 30
* Looks for columns in the csv named `assertion{x}`
* If an assertion is not found in the csv, it is not processed (for example, `assertion15` was not found)
* The assertions are already the last columns in the csv, so this shouldn't be an issue.

### DIFF
--- a/scripts/create-tests.js
+++ b/scripts/create-tests.js
@@ -364,7 +364,10 @@ function createTestFile (test, refs, commands) {
   let references  = getReferences(refs.example, test.refs);
   addSetupScript(test.setupScript, setupFileName);
 
-  for (let i=1; i<6; i++) {
+  for (let i=1; i<31; i++) {
+    if (!test["assertion"+i]) {
+      continue;
+    }
     addAssertion(test["assertion"+i]);
   }
 


mfairchild365:
@jongund can you review this too?

jongund:
@mfairchild365 
Have you tested it to see if the assertions greater than 5 actually show up in the test harness?
The original reason for this limit was a restriction in the test harness of 5 assertions, but I know this was discussed and was suppose to be fixed in the harness.

mfairchild365:
@jongund Yes, I tested it and it shows as expected in the harness.

There are no limits on the number of assertions in the harness. For example, see

- https://github.com/w3c/aria-at/blob/master/tests/resources/aria-at-harness.mjs#L256
- https://github.com/w3c/aria-at/blob/master/tests/resources/aria-at-harness.mjs#L310

`create-tests.js` is the only file that has a limit on the number of assertions (see PR diff). So, unless I'm missing something, I don't think this issue resides in the test harness.

PR #267, opened 2020-06-05
Fix issue where menubar test 15 for jaws was not rendering

mfairchild365:
The jaws task name was incorrect
fixes #230

Issue #230, opened 2020-05-29
Tester issue report for: "Navigate to checked menu item radio in interaction mode"


mfairchild365:
### Test file at exact commit

[tests/menubar-editor/test-15-navigate-to-checked-menuitemradio-in-submenu-interaction.html](https://github.com/w3c/aria-at/blob/2a3c46f1cbd20e7f2f0a706052d2d82aaf207a29/tests/menubar-editor/test-15-navigate-to-checked-menuitemradio-in-submenu-interaction.html)

### Cycle:

Test Pilot (2020-05-27)

### AT:

JAWS (version 2020.2004.66 - April 2020)

### Browser:

Chrome (version 83.0.4103.61)

### Description

This test is not rendering. I get an empty page.

mfairchild365:
@mcking65 I created a PR to fix an issue with the test data that was causing this. As I understand it, we can't deploy that fix to the prod runner for this test run since this is an issue with the test data itself. So this test will have to remain unfinished. In other words, it is impossible for me to complete this test run. 

@zcorpan does that sound right?

PR #272, opened 2020-06-09
validate that a command actually exists for the combination of task, mode, and AT

mfairchild365:
Should help to prevent errors similar to #230. Related to #267


mfairchild365:
@jongund can you review this?

PR #303, opened 2020-10-19
Install pr-previewer to make it easier to review tests

rmeritz:
The [pr-preview](https://github.com/tobie/pr-preview) bot is already installed on all w3c repos. It automatically generates a link to the rendered raw html index page for a branch such that it is easier to review changes on that branch. 

In order to use the bot, I added a new index file. This is just a table that links to the existing index page for each test and the existing review page for each test.

As users want to be able to review things on branches and not just in master I'm commiting the generated review files to all branches and not just the `gh-pages` branch. 

:warning: Adding this PR as a draft in order to test the installation of the bot. Don't merge yet.


rmeritz:
The pr-preview bot doesn't seem to be generating any changes to the PR. I'm not sure of the root cause but suspect it needs to have the config/index file on master to work. 

For right now I'm just looking at https://raw.githack.com/w3c/aria-at/review-index/review-index.html to see what the generated preview link would look like.

nschonni:
I think the App needs to be installed https://github.com/apps/pr-preview

rmeritz:
> I think the App needs to be installed https://github.com/apps/pr-preview

I don't think so it should already be installed on all `w3c` repos. See the following from the link you sent.

> Note that the following orgs already have blanket installs of pr-preview, which means that adding a config file is all you need to be setup if your repository is hosted in one of them:

> github.com/w3c

nschonni:
Sorry, you even said that in the description :(
Maybe it's because the file in `  "src_file": "test/checkbox/index.html",` is missing the "s" in "tests"

spectranaut:
I enabled the branch on my fork so you can see the "preview" link on the this PR, but the links from the index page do not work and I'm not quite sure why: https://github.com/spectranaut/aria-at/pull/7

evmiguel:
@spectranaut The links are pointing to an S3 location, which doesn't have public access.

spectranaut:
I'm merging this PR, we will fix the links in a follow up!

PR #306, opened 2020-10-21
Replace select with checkbox

evmiguel:
Replace select options for undesirable behavior with checkboxes.

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -366,19 +366,20 @@ Were there additional undesirable behaviors? <span class="required">(required)</
   <label for="problem-${c}-select">Select all undesirable behaviors <span class="required">(required: select at least one or mark "No additional undesirable behaviors")</span>:</label>
 </div>
 <div>
-  <select multiple name="problem-${c}-select" id="problem-${c}-select">
+  <fieldset> 
 `;
-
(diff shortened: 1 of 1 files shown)


spectranaut (changes requested):
When testing these changes, I noticed two weird behaviors:

1. This should be changed in this PR, @evmiguel:  If you click `Yes, there were additional undesirable behaviors`. but then don't select any of the undesirable behaviors, then I think you should get better validation errors. Right now it just highlights the box. It should probably add "required" but the "Select all undesirable behaviors :"

1. This is a UX question and I'm not sure needs to be fixed here. It seems a bit weird to make the checkboxes disabled unless `Yes, there were additional undesirable behaviors` is checked.  You might try to check a undesirable behavior checkbox (like i did) but it doesn't work until you click "yes, there were additional undesirable behaviors".

I think if you check an undesirable behavior, it should just switch the `No, there were no additional undesirable behaviors.` to `Yes, there were additional undesirable behaviors`.

I think this could be done in a follow up, maybe after user feedback -- I'm just not sure how to prioritize the change. Maybe @s3ththompson has an idea?

spectranaut (approved):
Actually, both of my concerns can be fixed in a follow up because other work depends on this. I made  "to triage" issues in the bocoup asana board for them.

PR #307, opened 2020-10-21
Switch from pr-previwer to github action preview link

rmeritz:
[Preview Tests](https://raw.githack.com/w3c/aria-at/pr-previewer-fixes/index.html)

The installed pr-previewer had two bugs:
1. The preview wasn't automatically attached to diffs unless they changed the number of tests.
2. Links to other pages from the review index got an access denied error.

I tried a couple of approaches to fixing including [reaching out to pr-previwer](https://github.com/tobie/pr-preview/issues/83) to find a solution before ultimately deciding to use a different solution and uninstall pr-previewer.

I switched to using a GitHub action called [pr-updater](https://github.com/tzkhan/pr-update-action) that adds templated branch name information to the title and body of all prs. 

The "Preview Tests" link above was automatically added in via the action. It uses the service rawgit, which pr-previewer provided a wrapper around, to preview link to the test index on the branch.

I created [a fork](https://github.com/rmeritz/pr-update-action) of the original action so that I didn't automatically need to include the branch name in the title. I've opened a [PR](https://github.com/tzkhan/pr-update-action/pull/16) to merge my changes into the original action.

I also updated the index file to show and link to the latest commit where the test example was changed.

### DIFF
--- a/.github/workflows/update-pr.yml
+++ b/.github/workflows/update-pr.yml
@@ -6,11 +6,11 @@ jobs:
   update_pr:
     runs-on: ubuntu-latest
     steps:
-    - uses: tzkhan/pr-update-action@v1.1.1
+    - uses: rmeritz/pr-update-action@930be7f0caa6f2e6339756fe2ef5b8b8963454cf
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"                   # required - allows the action to make calls to GitHub's rest API
-        branch-regex: '.*'                                           # required - regex to match text from the head branch name
-        title-template: '%branch%'                                  # required - text template to update title with
+        branch-regex: '.*'                                          # required - regex to match text from the head branch name
+        title-template: ''                                          # required - text template to update title with
         replace-title: false                                        # optional - whether to prefix or replace title with title-template
         uppercase-title: false
         body-template: '[Preview Tests](https://raw.githack.com/w3c/aria-at/%branch%/index.html)' # required - text template to prefix body
--- a/.github/workflows/update-pr.yml
+++ b/.github/workflows/update-pr.yml
@@ -1,4 +1,4 @@
-name: "Update Pull Request"
+name: "Update Pull Request with Preview Link"
 
 on: pull_request
 
--- a/.github/workflows/update-pr.yml
+++ b/.github/workflows/update-pr.yml
@@ -11,5 +11,5 @@ jobs:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
         head-branch-regex: '.*'
         title-template: ''
-        body-template: '[Preview Tests](https://raw.githack.com/w3c/aria-at/%branch%/index.html)'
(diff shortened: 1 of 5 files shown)


.github/workflows/update-pr.yml
```diff
@@ -0,0 +1,15 @@
...
+jobs:
+  update_pr:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: tzkhan/pr-update-action@v2
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
+        head-branch-regex: '.*'
+        title-template: ''
+        body-template: '[Preview Tests](https://raw.githack.com/w3c/aria-at/%branch%/index.html)'
```
spectranaut, line 14: I think you mean `%headbranch%`? The link doesn't work with `%branch%`

index.html
```diff
@@ -34,25 +32,32 @@ <h1>Test Patterns for Review</h1>
...
-        <th>Number of Tests</th>
+        <th># of Tests</th>
+        <th>Commit of Last Change</th>
       </tr>
           <tr>
             <td>checkbox</td>
-            <td><a href="./tests/checkbox/index.html">Index</a></td>
-            <td><a href="./review/checkbox.html">Review</a></td>
+            <td><a class="githack-link" href="./tests/checkbox/index.html">Index</a></td>
+            <td><a class="githack-link" href="./review/checkbox.html">Review</a></td>
```
spectranaut, line 41: What is this class for?

scripts/test-reviewer.mjs
```diff
@@ -170,7 +170,15 @@ for (let pattern in allTestsForPattern) {
...
-  patterns: Object.keys(allTestsForPattern).map(pattern => ({ name: pattern, numberOfTests: allTestsForPattern[pattern].length }))
+  patterns: Object.keys(allTestsForPattern).map(pattern => {
+    const lastCommit = spawnSync('git', ['log', '-n1', '--oneline', path.join('.', 'tests', pattern)]).stdout.toString();
+    return {
+      name: pattern,
+      numberOfTests: allTestsForPattern[pattern].length,
+      commit: lastCommit.split(' ')[0],
+      commitDescription: lastCommit
+    }
+  })
```
spectranaut, line 181: I guess we only needed this for pr-previewer, but maybe it's nice to leave in.

spectranaut (changes requested):
The link doesn't work in the description, I think you used the wrong variable name!

scripts/test-reviewer.mjs
rmeritz: I think it's still a general improvement to the page.

index.html
rmeritz: I added it in for an attempted fix where I added a script that targeted this class. Removed it now.

.github/workflows/update-pr.yml
rmeritz: Opps. This weekend my PR got merged into the action and they did a v2 release which changed the entire action and I just swtiched to it. Good catch. Fixed now.

spectranaut (approved):
awesome! :)

PR #308, opened 2020-10-21
Save partial results

evmiguel:
This PR is dependent on #306. 

Note: I am not on the project tomorrow, 10/22. When #306 is merged, please rebase this branch on top of master.

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -420,6 +420,11 @@ Were there additional undesirable behaviors? <span class="required">(required)</
       submitResult();
     });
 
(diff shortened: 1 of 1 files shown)


spectranaut (changes requested):
Having reviewed the other PR, I don't think this change is necessary as there is really nothing to get from the iframe -- this test harness doesn't need to do anything or supply anything when we save a partial result -- so this event listener is essentially a no-op. If we wanted, we could support a concept of a partial result so that we didn't need to serialize the form data (so the results object would have incomplete data, and the harness would know how to accept a results object and re-populate the form), but that would be a much bigger change and also probably not worth it.

I'm going to close this PR.

PR #310, opened 2020-10-22
Add github action to commit generated review files

rmeritz:
[Preview Tests](https://raw.githack.com/w3c/aria-at/github-actions/index.html)

- Added a github action to automatically generate all test and review files
- Added a new script that generates all test files and not just one example at at time
- Removed existing generated review and test files for testing purposes and regenerated them

References:
https://github.com/marketplace/actions/git-auto-commit
https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-nodejs

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/aria-at/pull/310.html" title="Last updated on Oct 22, 2020, 6:23 PM UTC (17f5dc2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/aria-at/310/98230de...17f5dc2.html" title="Last updated on Oct 22, 2020, 6:23 PM UTC (17f5dc2)">Diff</a>


review/checkbox.html
```diff
@@ -74,7 +74,7 @@ <h2>Test 1: Navigate to an unchecked checkbox in reading mode</h2>
       <ul>
         <li>Mode: reading</i>
         <li>Applies to: jaws, nvda</li>
-        <li>Lasted edited: Wed May 13 19:00:23 2020 -0500
```
spectranaut, line 77: I hope this is not updated for every test file every PR because that would be pretty obnoxious..

edit: looks like it's not, according to the code

spectranaut (approved):
This looks great! :)

spectranaut:
@rmeritz looks like html2 parser was accidentally taken out when removing the app, that is the failure here! Can you push a fix?

rmeritz:
> @rmeritz looks like html2 parser was accidentally taken out when removing the app, that is the failure here! Can you push a fix?

Fixed. I also switched the action to use `npm` instead of `yarn` which we don't seem to be using for this project.

PR #316, opened 2020-10-29
fix: Missing end li

nschonni:


nschonni:
@spectranaut I wasn't sure if the generated index should be included or not

spectranaut (approved):
It should, and thanks @nschonni ! :)

PR #329, opened 2020-11-10
chore: Skip double build on Dependabot linting

nschonni:
Only submitted through PRs, so it doesn't need the push build


nschonni:
Rolled this into https://github.com/w3c/aria-at/pull/632

Issue #305, opened 2020-10-20
Document Plan to Change the Generated Test Files


rmeritz:
- In order to support being able to preserve test results over different test versions, we need to be able to diff the test file against a test file in a different version to see if the tests have semantically changed or not. Currently diffing the test file is not meaningful because so much of the test is actually containing in other referenced files.

- The csv format for the test author will stay the same. Only the `create-tests.js` script will be modified. A new script will also be provided to test authors in order to allow them to update the `reference` example code from `aria-practices` in a structured/versioned.

- Details of the new format can be seen in this annoted example file of the new test file format: https://github.com/w3c/aria-at/compare/spike-one-file

PR #330, opened 2020-11-10
Create tests for APG design pattern example: Tri-State Checkbox

jscholes:
[Preview Tests](https://raw.githack.com/w3c/aria-at/tests/checkbox-tri-state/index.html)

Resolves #322.

PR #333, opened 2020-11-16
Change format of generated test files to be comparable

rmeritz:
[Preview Tests](https://raw.githack.com/w3c/aria-at/test-html/index.html)

See the annotated file in this issue for more infomation about the reason for this change: https://github.com/w3c/aria-at/issues/305

* Inline the relevant script and remove scripts.js
* Inline all the test data and remove the test json file
* Inline the relevant command and remove commands.json


tests/checkbox/test-01-navigate-to-unchecked-checkbox-reading.html
```diff
@@ -6,18 +6,81 @@
 <link rel="help" href="https://w3c.github.io/aria/#checkbox">
 <link rel="help" href="https://w3c.github.io/aria/#aria-checked">
 
-<script src="scripts.js"></script>
+<script>
+  var scripts = {
```
spectranaut, line 10: this is a bit strange, it might be nice to not include it if it's not necessary. but because they are generate files I suppose it doesn't matter, and this is not blocking.

spectranaut (approved):
This is works great, and the changes are very simple and straightforward.

jscholes:
@rmeritz or @spectranaut, as a test developer, is there anything I have to do to comply with these changes for PRs that haven't yet been merged?  E.g. https://github.com/w3c/aria-at/pull/330.

rmeritz:
@jscholes - Nope these changes don't affect anything about the test writing. Once this PR is merged when you regenerate the generated files will have a different format but you don't need to worry about it at all. Thanks for checking though.

PR #339, opened 2020-11-20
Add script to pull in updated references from aria-practices

rmeritz:
[Preview Tests](https://raw.githack.com/w3c/aria-at/update-reference/index.html)

* Prior to this PR examples were manually copied into the reference files from the https://github.com/w3c/aria-practices repo. This means there might be drift from the examples to what is actually under test. We don't want to actually just pull them directly from the aria-practices repo because if an example changes the test itself might need to change. It is important for the test writer to have control of when to pull in updates. 
* When running the tests, it is important to be able to diff versions of test files and know if a change has accused that would need would cause the tests to be rerun. See https://github.com/w3c/aria-at/issues/305 for more details. To serve this purpose a timestamps subdirectory provides a form of version control for the reference. 

The script added in this PR makes the process of updating the reference files less manual. The files get pulled into the repo automatically. However, for now, I'm still having the test writer commit the new reference files themselves and update the `data/reference.csv` file. This is to avoid problems with dirty working directories and to give the test writer more control over when the update should happen. The script provides instructions to the user for making the update. 

Once this merged I will update the wiki with documentation about it. 

Tests are written in `aria-at` for specific examples in the `aria-practices` repo. To clarify this relationship, I'm normalizing the names of the examples in this repo to match the exact names in `aria-practices`. `combobox-autocomplete-both` and `menubar-editor` already match the names of specific examples but `checkbox` didn't. There are actually two checkbox examples `checkbox-1` and `checkbox-2` so I changed the name of `checkbox` to match. I think this is a standard we should enforce going forward.


jscholes:
> I'm normalizing the names of the examples in this repo to match the exact names in aria-practices. combobox-autocomplete-both and menubar-editor already match the names of specific examples but checkbox didn't. There are actually two checkbox examples checkbox-1 and checkbox-2 so I changed the name of checkbox to match. I think this is a standard we should enforce going forward.

I agree with the intention behind this on the ARIA-AT side, but checkbox2 is a terrible name for a pattern example on the APG front.  When used in a table for review purposes, that matters.  I filed an issue at https://github.com/w3c/aria-practices/issues/1630, but maybe it would be a good idea to allow tester writers to:

1. override the directory name for a test plan if needs be; and
2. specify the corresponding APG example name somewhere else?

jscholes:
@rmeritz Does this also pull in the corresponding CSS and JS?  If not, what would the process look like for updating those after manual import?

I'd love to have a kind of before/after example of process with and without this script available.

rmeritz:
> Does this also pull in the corresponding CSS and JS? If not, what would the process look like for updating those after manual import?

@jscholes - Yes, this also pulls in any other assets Js, Css or images automatically.

> override the directory name for a test plan if needs be; and
> specify the corresponding APG example name somewhere else?

@jscholes - I will add a way of specifying the APG directory name without forcing users to have it be an exact match.

jscholes:
@rmeritz When I run this on Windows, I get an error:

```
>npm run update-reference combobox-select-only

> aria-at@1.0.0 update-reference L:\Dropbox\Dev\aria-at
> node ./scripts/update-reference.js "combobox-select-only"

Cloning the aria-practice repo.
Locating the matching example file combobox-select-only.html
Coping assets to timestamped local directory.
(node:14004) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'split' of undefined
    at copyExampleToRepo (L:\Dropbox\Dev\aria-at\scripts\update-reference.js:79:82)
(node:14004) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:14004) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Not sure if I'm doing something wrong?  I'm on Node v12.19.1.

I used the select-only combobox example because there are multiple HTML files in a single directory, all corresponding to different combobox implementations.  Likewise, within the `css` and `js` dirs, there is a one-to-one mapping between an asset and pattern, so e.g. only `js/select-only.js` is required for the select-only example.

This differs from something like checkbox, where each checkbox example has its own directory with an HTML file and `js` subdir.  Although even there, the `css` directory is a sibling of the two checkbox dirs and shared across the two.  Finally, not all examples use the `id="ex1"` pattern for the container div.

For maximum automation effectiveness, I think the APG would need to be significantly reorganised to follow a consistent structure; it might be good to discuss on an ARIA-AT call and ultimately come up with that set of recommendations, and document the process so that test writers can achieve the same thing manually in a consistent fashion.

rmeritz:
@jscholes - Thanks so much for trying this out and providing detailed feedback.

> @rmeritz When I run this on Windows, I get an error:

I wasn't handling Windows-style paths before which was probably the source of your error. It would be great if you tried it again. I added more error handling as well. 

> I used the select-only combobox example because there are multiple HTML files in a single directory, all corresponding to different combobox implementations. Likewise, within the css and js dirs, there is a one-to-one mapping between an asset and pattern, so e.g. only js/select-only.js is required for the select-only example.

> This differs from something like checkbox, where each checkbox example has its own directory with an HTML file and js subdir. Although even there, the css directory is a sibling of the two checkbox dirs and shared across the two. 

The way I'm dealing with inconsistencies in the naming of assets in the aria-practices repo is just to copy over all the js/css assets for each example folder regardless of if they are used in a specific example or not. So I don't think it will be a problem coping over that assets regardless of how the subdirs are organized. Though this script could certainly be made nicer if aria-practices was more consistent. 

> Finally, not all examples use the id="ex1" pattern for the container div.

Yes, I looked at it seems like 90% use "ex1", but not 100%. This is a major impediment to full automation.  I changed the script such that the `html` is copied in as is with all the extra info and provide I provide instructions to the test writer to remove it and save the file. The process will need to remain 50% manual until apg-practices is cleaned up. 

> For maximum automation effectiveness, I think the APG would need to be significantly reorganized to follow a consistent structure; it might be good to discuss on an ARIA-AT call and ultimately come up with that set of recommendations, and document the process so that test writers can achieve the same thing manually in a consistent fashion.

I will talk with my team about moving this forward so that in the future this script could be improved. I still want to add some sort of script now to ensure that when the reference is copied in it is copied into a dated folder to provide a mechanism of "version control" for the tests.

jscholes:
@ rmeritz 

> I wasn't handling Windows-style paths before which was probably the source of your error. It would be great if you tried it again. I added more error handling as well.

Thanks; it works now on Windows.  It does throw an unhelpful error when references.csv exists but is empty or invalid, so that could be a good case to check for:

```
(node:2920) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'split' of undefined
    at copyExampleToRepo (L:\Dropbox\Dev\aria-at\scripts\update-reference.js:73:88)
    at Object.<anonymous> (L:\Dropbox\Dev\aria-at\scripts\update-reference.js:115:1)
    at Module._compile (internal/modules/cjs/loader.js:1015:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1035:10)
    at Module.load (internal/modules/cjs/loader.js:879:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47
```

> I still want to add some sort of script now to ensure that when the reference is copied in it is copied into a dated folder to provide a mechanism of "version control" for the tests.

I agree, but think it's critical that this pattern is also documented somewhere so that a test writer can achieve the same thing manually.  E.g. I'm on a metered internet connection and already have an up-to-date clone of https://github.com/w3c/aria-practices, so the fact that the script re-clones it afresh each time (I think?) is something I want to avoid.

This could be as simple as updating the [How to contribute tests](https://github.com/w3c/aria-at/wiki/How-to-contribute-tests) page to indicate the date/time format for the directory name, etc.  Having said that, the information there about references.csv currently seems incomplete anyway, e.g. it doesn't mention the `author`, `authorEmail`, `example` or `reference` fields.

rmeritz:
@jscholes - Thanks again for the feedback. Sorry it took me a long time to getting around to addressing it. 

>  It does throw an unhelpful error when references.csv exists but is empty or invalid, so that could be a good case to check for:

I added error handling here.

> I agree, but think it's critical that this pattern is also documented somewhere so that a test writer can achieve the same thing manually. E.g. I'm on a metered internet connection and already have an up-to-date clone of https://github.com/w3c/aria-practices, so the fact that the script re-clones it afresh each time (I think?) is something I want to avoid.

I updated the `reference.csv` section of the wiki on how to contribute tests to include more information about this here: https://github.com/w3c/aria-at/wiki/How-to-contribute-tests#refid

I've also added an optional pararmeter to the script `-r` which allows the user to pass in their local directory for their `aria-practices` clone and skip the download.

jscholes:
@rmeritz 

> I've also added an optional pararmeter to the script -r which allows the user to pass in their local directory for their aria-practices clone and skip the download.

Fantastic, thanks for that.

jongund:
I think many of the APG example file names need to be improved to be more descriptive.
This is happening as examples are updated or new one are added.
But I think the ARIA-AT should rename files to be more be more maningful.
Jon

Issue #349, opened 2020-11-26
Automation test format


zcorpan:
Following discussion in https://docs.google.com/document/d/1jDf_gEQjRppLyEDEW0qPVbq1QaCT36bVDl9JkI-nvCU/edit# (also see #337)

cc @sinabahram @jscholes

We agreed that, to enable automation:

- Tests need to provide full and exact instructions for which keys to press
- Screen readers give some output in response to a keypress, and we generally want to assert multiple things. We can do that by having specialized assertions that look for how many times a specific substring appears (for example, for `assert_accname: Lettuce`, pass if the substring "Lettuce" appears exactly once). And also, we want to assert that the full string matches an expected string.

For example, for the test "Navigate to an unchecked checkbox in reading mode" for NVDA, I envision a text file that acts as the test with the full sequence of where to navigate the browser to, which keys to press, and the assertions. It could look something like this:

```
nav: reference/two-state-checkbox.html
press: h
assert_contains: Sandwich Condiments
press: x
assert_role: checkbox
assert_checked: false
assert_accname: Lettuce
assert_equals: Sandwich Condiments grouping list with 4 items Lettuce check box not checked
press: x
assert_accname: Tomato
press: Shift+x
assert_role: checkbox
assert_checked: false
assert_accname: Lettuce
assert_equals: Lettuce check box not checked
press: Shift+h
assert_contains: Sandwich Condiments
press: f
assert_role: checkbox
assert_checked: false
assert_accname: Lettuce
assert_equals: Sandwich Condiments grouping list with 4 items Lettuce check box not checked
press: f
assert_accname: Tomato
press: Shift+f
assert_role: checkbox
assert_checked: false
assert_accname: Lettuce
assert_equals: Lettuce check box not checked
press: Shift+h
assert_contains: Sandwich Condiments
press: Tab
assert_role: checkbox
assert_checked: false
assert_accname: Lettuce
assert_equals: Sandwich Condiments grouping list with 4 items Lettuce check box not checked
press: Tab
assert_accname: Tomato
press: Shift+Tab
assert_role: checkbox
assert_checked: false
assert_accname: Lettuce
assert_equals: Lettuce check box not checked
```

Note that this isn't the format I think we should *write* tests in, that can still be csv and avoid repetition, but the above could be generated from the csv source.

In addition, the NVDA project could choose to have "test expectations" files alongside these, that give the actual output after each interaction and/or each assertion.

```
nav: reference/two-state-checkbox.html
[...]

PR #350, opened 2020-12-02
Add a prototype automated test file (not generated)

zcorpan:
See https://github.com/w3c/aria-at/issues/349

---

This isn't ready to be merged, but creating a PR to facilitate review and discussion.


tests/checkbox/automated/test-01-navigate-to-unchecked-checkbox-reading.nvda.json
```diff
@@ -0,0 +1,244 @@
+[
+  {
+    "nav": [
+      "reference/two-state-checkbox.html"
```
rmeritz, line 4: This has moved btw.

tests/checkbox/automated/test-01-navigate-to-unchecked-checkbox-reading.nvda.json
```diff
@@ -0,0 +1,244 @@
...
+    "clear_output": []
+  },
+  {
+    "press": [
+      "x"
+    ]
+  },
+  {
+    "assert_role": [
+      "checkbox"
```
rmeritz, line 23: If there is going to be a test file for each AT, do is maybe make more sense for the normalization of role from `checkbox` to `check box` to live here? (I'm not sure...)

zcorpan: I think it's better to have the mapping in the AT's repo (if open source, at least), so that the test can be applicable across versions of the AT even if they changed how a role is conveyed. If it's in `aria-at` then it makes things annoying both for the AT vendor and the aria-at CG when it changes (though may be rare, but still).

zcorpan:
`update_pr` gets confused by this PR it seems.

jscholes:
@zcorpan 

> I think it's better to have the mapping in the AT's repo (if open source, at least)

As indicated on a previous call, I don't agree with this.  It creates fragmentation because NVDA is the only truly open source AT that we're likely to support in the short to medium term future.  Even TalkBack on Android can't be considered open source in the way NVDA is, because the source is only made available along with public releases.

You wrote:

> If it's in aria-at then it makes things annoying both for the AT vendor and the aria-at CG when it changes

Why does it make things annoying for the AT vendor?  If they are using some part of ARIA-AT as a dependency in their build toolchain, they should be expected to update it along with other similar dependencies.  If they feel that they are blocking on someone on the ARIA-AT side making a role mapping or other update, they can submit a PR or file an issue.

Same goes for the ARIA-AT side.  With this proposal, the ARIA-AT project would be responsible for maintaining and updating the large majority of role mappings, but not all of them.  It may be rare for existing mappings to change, but new roles may be invented which are implemented in multiple ATs at around the same time.  At that point, somebody from the ARIA-AT CG (or whoever) could get stuck in updating different AT mappings, only to find that, say, NVDA's mappings were maintained elsewhere and hadn't been updated yet.  At which point I'd fully expect them to at least consider submitting a PR against NVDA, which would mean they'd done the work anyway.

At the most basic level, having NVDA's role mappings be maintained by NVAccess and having all others be maintained by the ARIA-AT CG is confusing.  Not to mention the fact that, AFAIK, NVAccess haven't agreed to take on that responsibility, so it may come down to somebody else to take it on anyway.  At which point we'd just be spreading things across repos for no gain.

PR #351, opened 2020-12-02
Add aria practice guide links to reference.csv(s)

evmiguel:
[Preview Tests](https://raw.githack.com/w3c/aria-at/aria-practice-link-reference-csv/index.html)

### DIFF
--- a/tests/checkbox/data/references.csv
+++ b/tests/checkbox/data/references.csv
@@ -2,6 +2,7 @@ refId,value
 author,Michael Fairchild
 authorEmail,michael.fairchild@deque.com
 reference,reference/2020-11-23_175030/checkbox-1/checkbox-1.html
+designPattern,https://w3c.github.io/aria-practices/#checkbox
 example,https://w3c.github.io/aria-practices/examples/checkbox/checkbox-1/checkbox-1.html
 checkbox,https://w3c.github.io/aria/#checkbox
 group,https://w3c.github.io/aria/#group
--- a/tests/menubar-editor/data/references.csv
+++ b/tests/menubar-editor/data/references.csv
@@ -14,3 +14,4 @@ aria-haspopup,https://w3c.github.io/aria/#aria-haspopup
 aria-expanded,https://w3c.github.io/aria/#aria-expanded
 aria-disabled,https://w3c.github.io/aria/#aria-disabled
 aria-label,https://w3c.github.io/aria/#aria-label
+designPattern,https://w3c.github.io/aria-practices/#menu
--- a/index.html
+++ b/index.html
@@ -40,7 +40,7 @@
             <td><a href="./tests/checkbox/index.html">Index</a></td>
             <td><a href="./review/checkbox.html">Review</a></td>
             <td>26</td>
-            <td><a href="https://github.com/w3c/aria-at/commit/ec29881" target="_blank">ec29881 Add ability to use a local aria-repo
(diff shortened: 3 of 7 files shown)


jscholes:
@evmiguel I'm not sure I understand the intent behind/utility of this.  Using the checkbox tests as an example, there is an "example" row with the URL: https://w3c.github.io/aria-practices/#checkbox.  But this is wrong; that URL is the link to the design pattern in the APG, not the example being tested.  Therefore, during development of new pattern tests, we've been using that row for the actual example link, the one that in this PR is called "practiceGuide".

In other words:

1. These seem back-to-front.  If we need rows for the APG section and example, they should be reversed.
2. Do we need both?  The example will always link to the APG section.
3. If we need both, maybe "pattern", "designPattern" or similar is clearer than "practiceGuide"?

Let me know your thoughts.

evmiguel:
@jscholes Thank you for taking a look at this. We need both the `example` value and what I have called `practiceGuide` for ARIA-AT app, as we want to link to both the example and the documentation around the pattern. Given your feedback, I will update the key to `designPattern`.

rmeritz:
@evmiguel - This looks good can you update the "How to write tests" to add this as a `refId` before merging?

Also, @jscholes wanted to make sure you are aware that we are adding another field to the list of fields that should be added the the `references.csv`. It is called `practiceGuide` it links to the high-level design pattern for an example. Like https://www.w3.org/TR/wai-aria-practices-1.1/#checkbox. (This URL is already referenced in most `refernce.csv` files but not in a way that makes it clear across examples what is the most relevant high-level link is.

This change is being made to enable us to grab that data and display it in a new aria-at-app reports page.

jscholes:
@rmeritz 

> It is called practiceGuide it links to the high-level design pattern for an example. Like https://www.w3.org/TR/wai-aria-practices-1.1/#checkbox.

Understood, thanks for the additional context.  The changes in `tests/checkbox/data/references.csv` are the wrong way round.  In the updated file, "designPattern" links to the example while "example" links to the pattern.

evmiguel:
> @rmeritz
> 
> > It is called practiceGuide it links to the high-level design pattern for an example. Like https://www.w3.org/TR/wai-aria-practices-1.1/#checkbox.
> 
> Understood, thanks for the additional context. The changes in `tests/checkbox/data/references.csv` are the wrong way round. In the updated file, "designPattern" links to the example while "example" links to the pattern.

@jscholes I noticed that, and I changed it!

tests/checkbox/data/references.csv
```diff
@@ -1,7 +1,8 @@
 refId,value
 author,Michael Fairchild
 authorEmail,michael.fairchild@deque.com
-reference,reference/2020-11-23_175030/checkbox-1/checkbox-1.html
```
rmeritz, line 4: @evmiguel - There is an actual problem you seem to have lost my changes because I think you need to rebase. This needs to not change.

tests/menubar-editor/data/references.csv
```diff
@@ -1,8 +1,9 @@
 refId,value
 author,Jon Gunderson
 authorEmail, jongund@illinois.edu
-reference,reference/2020-11-23_175528/menubar-editor.html
```
rmeritz, line 4: Same problem.

jscholes:
Also, in `tests/combobox-autocomplete-both/data/references.csv`, "authorEmail" is mistyped as "authofEmail".  Not sure if that's in scope to fix here, given that:

1. it presumably wasn't introduced in this PR; and
2. these tests will most likely be replaced with an updated version anyway.

PR #359, opened 2020-12-17
update scripts to only check first file and save last edited date fro…

dawnwages:
[Preview Tests](https://raw.githack.com/w3c/aria-at/last-edited-test-reviewer/index.html)

…m important files == git log from data and reference dirs

The code could be a little cleaner but I was able to get the optimal case complete instead of the the `test.csv`
https://app.asana.com/0/1193055321453706/1199152269152973/f

**More notes on the ticket requirements:**
- only the data and reference folders but - git log 
- last file was changed ==> Any file in relevant directory related files
- once for every test 
- can you check all of the underlying file, then just check test.csv and add a "~"
- alternative: instead of checking if the test file is changed, the most relevant stuff is in test.csv

### DIFF
--- a/scripts/review-template.mustache
+++ b/scripts/review-template.mustache
@@ -113,7 +113,7 @@
       <ul>
         <li>Mode: {{mode}}</li>
         <li>Applies to: {{allReleventATsFormatted}}</li>
-        <li>Lasted edited: {{lastEdited}}</li>
+        <li>Last edited: ~ {{lastEdited}}</li>
         <li>Tests:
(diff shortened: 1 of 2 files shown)


dawnwages:
better/future solution: for each important dir in list, push lastDirEdited date to array, then pick latest date as lastEdited
short on time, but happy to go back and update code to the above ^

dawnwages:
`npm run review-tests`

s3ththompson:
Superseded by #451

PR #361, opened 2020-12-17
Create tests for APG design pattern example: Modal Dialog

### LINKED ISSUE #348: Create tests for APG design pattern example: Modal Dialog
#### Applies To

* APG design pattern: [Dialog (Modal)](https://w3c.github.io/aria-practices/#dialog_modal)
* Specific example: [Modal Dialog](https://w3c.github.io/aria-practices/examples/dialog-modal/dialog.html)

#### Testing Notes

This issue thread will house all documentation relating to the development of a test plan for the "Modal Dialog" APG example, which conforms to the "Dialog (Modal)" design pattern.  All follow-ups (including the test plan itself) will be added as new comments to facilitate notifications, but this initial comment will be updated with relevant details and URLs as needed for convenience.

#### Additional References

* [dialog (role)](https://w3c.github.io/aria/#dialog)
* [aria-modal (property)](https://w3c.github.io/aria/#aria-modal)
* [aria-describedby (property)](https://w3c.github.io/aria/#aria-describedby)

jscholes:
[Preview Tests](https://raw.githack.com/w3c/aria-at/tests/modal-dialog/index.html)

Resolves #348


IsaDC (approved):

jongund:
For test "Testing task: Navigate to the end of a modal dialog in reading mode"
The keyboard shortcut for JAWS to move to last form filed should change from control+end to insert+control+end

jscholes:
@jongund 

> For test "Testing task: Navigate to the end of a modal dialog in reading mode"
> The keyboard shortcut for JAWS to move to last form filed should change from control+end to insert+control+end

The aim of this test is just to move the reading/virtual cursor to the end of the dialog itself, and the assertions relate to the last thing in the dialog that users will encounter.  Here, it's the Cancel button.  The aim isn't specifically to jump to the last form field or other control type.

mcking65:
@jongund commented:

> For test "Testing task: Navigate to the end of a modal dialog in reading mode"
> The keyboard shortcut for JAWS to move to last form filed should change from control+end to insert+control+end

Agree with @jscholes.

Another way to think of this is what happens if JAWS has a but and doesn't respect the boundary of the dialog and moves the reading cursor outside of it. If we use the command to move to last form field, the reading cursor will move to the cancel button and you would not detect the failure to respect dialog boundary. However, if you use ctrl+end, the reading cursor would move outside the dialog to the end of main content. The test needs to be able to capture that kind of bug.

mcking65:
I think we need a better way of testing if NVDA is in browse mode rather than pressing escape. Our current instruction is:

>1. Insure NVDA is in browse mode by pressing Escape. Note: This command has no effect if NVDA is already in browse mode. 

The note is not correct if the focus is in a widget that processes escape, such as a dialog or menu.

I'm not letting this stop merge of this PR. But, I do think we need to address this somehow so I've opened #382.

mcking65 (changes requested):
This plan is excellent!! I found only one copy/paste type of error.

Mistake in tester instructions in:
* Test 11: Navigate to the last focusable element in a modal dialog
* Instructions say: 2. With focus on the last focusable element inside a modal dialog, navigate to the first focusable element in the same dialog. using the following commands:
* Should say opposite

jscholes:
@mcking65 Thanks, good catched.  Fixed in https://github.com/w3c/aria-at/pull/361/commits/ceeb5f6c35e0f25ac6ce1dab16df79dfba28591f.

jongund (approved):

jongund (approved):
I agree with Matt test 11 seems to need to have some attention.  When you open the test page focus is on the second textbox and it seems like it should be on the first textbox.

jscholes:
@jongund 

> I agree with Matt test 11 seems to need to have some attention. When you open the test page focus is on the second textbox and it seems like it should be on the first textbox.

I can't reproduce this specifically, but the setup script code is a bit flaky when executed automatically.  As covered in #369 and https://github.com/w3c/aria-at/issues/370#issuecomment-760431167, I believe that we would prefer testers to run the setup script(s) automatically via a button provided on the page instead, to allow the browser and screen reader to settle first.

mcking65 (approved):
@jscholes thanks for fast fix!

PR #399, opened 2021-03-15
Create tests for APG design pattern example: Radio Group Using Roving tabindex

### LINKED ISSUE #380: Create tests for APG design pattern example: Radio Group Using Roving tabindex
#### Applies To

* APG design pattern: [Radio Group](https://w3c.github.io/aria-practices/#radiobutton)
* Specific example: [Radio Group Example Using Roving tabindex](https://w3c.github.io/aria-practices/examples/radio/radio.html)

#### Testing Notes

This issue thread will house all documentation relating to the development of a test plan for the "Radio Group Using Roving tabindex" APG example, which conforms to the "Radio Group" design pattern.  All follow-ups (including the test plan itself) will be added as new comments to facilitate notifications, but this initial comment will be updated with relevant details and URLs as needed for convenience.

#### Additional References

* [radiogroup (role)](https://w3c.github.io/aria/#radiogroup)
* [radio (role)](https://w3c.github.io/aria/#radio)
* [aria-checked (state)](https://w3c.github.io/aria/#aria-checked)

jscholes:
[Preview Tests](https://raw.githack.com/w3c/aria-at/tests/radiogroup-roving-tabindex/build/index.html)

Resolves #380.


robfentress:
This is for Task ID 2:  When using <kbd>Control</kbd>+<kbd>Option</kbd>+<kbd>Right</kbd> in VoiceOver/Safari to navigate to the first radio button, upon reaching the radio button, VoiceOver surfaces, "To select this option, press <kbd>Control</kbd>+<kbd>Option</kbd>+<kbd>Right</kbd>."  Does this count as "Good Output" for the assertion, "State of the radio button, 'unchecked', is conveyed"?  

When I navigate to the radio button when it is in the _selected_ state, VoiceOver tells me it is selected.  Therefore, users familiar with VoiceOver could assume that, when VoiceOver tells them how to select the option, it means that option is unselected.  

I lean towards counting this assertion as being "Good Output", but would be curious what others think.  How explicit does the screen reader need to be for its output to count as actually communicating a particular state?  Should we provide additional guidance about things like this to get more inter-rater reliability?  What would that guidance be?

robfentress:
Meta question: At this stage, are you simply wanting feedback on whether there are bugs in the code?  Should I leave off feedback about choice of tests or assertions are included?

robfentress:
What is the reason why the heading, "Pizza Crust", has its CSS `display` property set to `none`?  I didn't notice it was even there until I refreshed the page displaying the example, which caused the heading to be displayed.  I assume that means there is something in the setup script that hides it on purpose, but I wonder if lazy people like me might try to do all their testing in one go, refreshing the page to reset things, rather than closing the window and reopening it, which might result in inconsistencies.  I vaguely recall the group addressing this somehow previously, but just in case it hasn't been noticed as an issue, I wanted to mention it.

jscholes:
@robfentress 

> What is the reason why the heading, "Pizza Crust", has its CSS `display` property set to `none`?

In the mark-up, the heading is inside the group.  So we can't put the tester's default focus on a link after the heading, because then their focus would already be inside the group, and they wouldn't be able to test the group-related announcements.  Given that, and also the fact that we don't want the heading to interfere with any assertions passing or failing, we hide it.  We could also just remove it from the example.

> I wonder if lazy people like me might try to do all their testing in one go, refreshing the page to reset things, rather than closing the window and reopening it

In that case they would be doing the testing wrong :) We have reached consensus on how to address this going forward; the page will have a button to run the setup scripts instead of them being executed automatically.  Testers will need to reset the page and press that button between each command, so we'll be providing an explicit reset button to refresh the page as well but they can also just refresh it manually.

Issue #363, opened 2020-12-18
Support for command sequences in tests


jscholes:
Currently, the test format allows for multiple commands, but it is implied that these should be executed in isolation.  For example, pressing F to find a checkbox, returning to a known point on the page and then pressing X to find the same control.

this isn't sufficient in some cases, and multiple commands are required in a sequence instead.  Three categories that spring to mind are:

1. AT virtual buffer structure creating the need for multiple consecutive executions of the same command.  For example, JAWS requires multiple presses of Down Arrow to read the information about a table container as well as the text and coordinates of the first cell.  The same applies to VO+Left/Right Arrow when moving into certain components.
2. AT behaviour creating the need for one key to be pressed followed by another.  To reuse the table example from point `#1`, a JAWS tester will need to press T followed by Down Arrow to locate the table and move into the first cell.
3. Multiple commands needed in a certain time period/window.  For example, type-ahead using multiple characters in a listbox.

Note that for points `#1` and `#2`, the intermediate output is needed for testers to be able to make a verdict about the assertions.  For example after pressing T to find a table, testers should listen to the information about the table before pressing the next command to hear information about the first cell.  But for point `#3`, we generally only care about the speech output when the final destination is reached (e.g. the listbox option matching all typed characters).

Currently, this can be hacked via an entry in keys.mjs, e.g.:

`export const T_THEN_DOWN = "T followed by Down Arrow";`

This has multiple downsides, including:

* It's non-obvious for new test writers.  Even if we were to document it, it would still be a hack, and may introduce inconsistencies if different test authors used different wording.
* It makes the keys file, which is already in need of some organisation,  crowded.
* It's a blunt tool which doesn't really allow distinction between the three use cases outlined above.
* It will frustrate any future plans to change the wording for all command sequences, localise the testing UI, etc.

mfairchild365:
This could be relevant to automation. I bet that using an array of commands would be easier to process from an automation perspective.

jscholes:
Points of interest from the February 25, 2021 community group meeting:

* We don't want so-called "open-ended" sequences, where a human or automated tester is instructed to press a key multiple times until a condition is met.  Instead, all testing tasks should encode an explicit number of commands, even if it is the same command a set number of times.
* To that end, `commands.csv` needs to accept 1 or more commands in each individual column, as opposed to just one which is the current behaviour.
* When a task includes multiple commands in a sequence, assertions should be evaluated based on aggregate output from all of those commands.
[...]

PR #434, opened 2021-05-19
Replace harness instruction page rendering

mzgoddard:
The harness module is given a test json of commands to perform and assertions to record the results of. This change replaces the manual DOM rendering and manual updates of the test instructions and result form with an object representing the test results, from which a second object, the instruction document, is created representing the structure, text, and user event handlers to update the first object. This separates the application state of the instructions and result entry from the DOM so that it can be reused outside the harness module.

This change further uses the added instruction document type and produces the existing DOM with a small virtual dom module. From the perspective of someone running the commands and entering the results into the form, there should be no change in the page’s behavior.

## Progress

This change is almost complete. Briefly, I think to complete it, the result state needs to be turned into the summary object for submission. Below is an overview of the steps implemented.

- [x] Type definitions
  - [x] Define the type of the Test JSON input, stored in the harness `behavior` module scope variable
  - [x] Define the type of the Commands JSON input
  - [x] Define the type of the Support JSON input
  - [x] Define the type of the submitted results produced when Submit Results is clicked
  - [x] Define a type of the state of the user input results
  - [x] Define a type that represents the displayed text and result input fields of the test input and state of the results
- [x] Update state of results from user events sent to functions in the instruction document representation
  - [x] Update atOutput text for a command
  - [x] Update an assertion for a command
  - [x] Update an additional assertion for a command
  - [x] Update if there are or are not undesirable results
  - [x] Update individual undesirable entries
  - [x] Update other undesirable text input
  - [x] Update if field should highlight when required when validating
- [x] Create the document object following the defined type
- [x] Create the result submission following the defined type
- [x] Track document item focus when updating the result state
  - [x] Focus header after page load
  - [x] Focus next or previous checkbox when pressing relevant keyboard keys
  - [x] Focus first required and not set result input after validating before submitting the result
- [x] Render the document object into DOM
  - [x] Render updates with a virtual dom to maintain minimal changes to dom like prior manual change implementation
  - [x] Transform DOM events into relevant input to functions in document object
    - [x] Transform text input change event
    - [x] Transform radio click event
    - [x] Transform checkbox change and keydown event
    - [x] Transform open test pop click event
    - [x] Transform submit results click event

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -573,7 +573,11 @@ function renderVirtualInstructionDocument(doc) {
 
     instructAssertions(doc.instructions.assertions),
 
-    button(onclick(doc.instructions.openTestPage.click), rich(doc.instructions.openTestPage.button)),
+    button(
+      disabled(!doc.instructions.openTestPage.enabled),
+      onclick(doc.instructions.openTestPage.click),
+      rich(doc.instructions.openTestPage.button)
+    ),
 
     resultHeader(doc.results.header),
 
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -1,20 +1,25 @@
-import {commandsAPI} from './at-commands.mjs';
+import {commandsAPI} from "./at-commands.mjs";
+import {element, fragment, property, attribute, className, style, focus, render} from "./vrender.mjs";
+import {
+  AssertionResultMap,
+  CommonResultMap,
+  createEnumMap,
+  HasUnexpectedBehaviorMap,
+  TestRun,
+  UserActionMap,
+  userCloseWindow,
+  userOpenWindow,
+  WhitespaceStyleMap,
+} from "./aria-at-test-run.mjs";
 
-const UNDESIRABLES = [
+const UNEXPECTED_BEHAVIORS = [
   "Output is excessively verbose, e.g., includes redundant and/or irrelevant speech",
   "Reading cursor position changed in an unexpected manner",
   "Screen reader became extremely sluggish",
   "Screen reader crashed",
-  "Browser crashed"
+  "Browser crashed",
 ];
 
-const TEST_HTML_OUTLINE = `
-<div>
-  <section id='errors' style='display:none'><h2>Test cannot be performed due to error(s)!</h2><ul></ul><hr></section>
-  <section id='instructions'></section>
-  <section id='record-results'></section>
-</div>
-`;
 const PAGE_STYLES = `
   table {
     border-collapse: collapse;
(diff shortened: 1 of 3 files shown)


tests/resources/vrender.mjs
```diff
@@ -0,0 +1,906 @@
+/**
```
alflennik, line 1: This file is pretty mind blowing! This ain't everyday JS, this is shiny stuff. I find myself wondering how you did this!

I think, and correct me if I'm wrong, that this is a virtual DOM implementation to be used for rendering the tests, and I believe from our conversations that we wouldn't use it in the app, it would be confined to the aria-at repo. 

Which of course, the thought of having to come in here to fix something is making me sweat slightly, haha. It's frankly over my head right now. So my question is, is there a way to avoid having to maintain this ourselves? I'm sure the reasoning is solid but I was hoping you would take a little time to help me grasp it.

Finally, I was wondering if there's a good way to test this file specifically. How do I put it through its paces so I feel like I gave it the review it deserves?

tests/resources/aria-at-harness.mjs
```diff
@@ -162,10 +176,107 @@ function executeScriptInTestPage() {
...
+
+/**
+ * @typedef Support
+ * @property {AT[]} ats
+ * @property {{system: string[]}} applies_to
+ * @property {{directory: string, name: string}[]} examples
+ */
+
+/**
+ * @typedef {{[mode in ATMode]: {[assitiveTech: string]: string;};}} CommandsAPI_ModeInstructions
```
alflennik, line 193: Acknowledging that `assistiveTech` does sounds kind of cool, given that `at` is used throughout the project, including the repo name, I think just `at` (or maybe `atName` based on the context?) would be a better term to use.

tests/resources/aria-at-harness.mjs
```diff
@@ -162,10 +176,107 @@ function executeScriptInTestPage() {
...
+/**
+ * @typedef {([string] | [string, string])[]} CommandAT
+ */
+
+/**
+ * @typedef {{[atName: string]: CommandAT}} CommandMode
+ */
+
+/**
+ * @typedef Command
```
alflennik, line 205: The word "command" in ARIA-AT is in rough shape since we’ve tried to use it to refer to both keys / inputs and test instructions / assertions. It's used consistently throughout this repo to refer to keys / inputs. @s3ththompson @jesdaigle @zcorpan What should we do about this? I see two options.

1. We double down on the term command as it is used here and come up with another term for instructions / assertions. "Steps" is one that comes to mind.
2. We try to rename it to "input" which came up as a good term in a conversation within the CG, and in the future command will refer to test instructions / assertions.

Is there a third option I haven't considered?

I have kind of been trying to push us along option 2.

I think doing nothing and letting the terminology wither is a poor option, much worse than 1 or 2. And since this terminology is exposed in the shared APIs between ARIA-AT and ARIA-AT App, there is no good way to isolate the repos.

tests/resources/aria-at-harness.mjs
```diff
@@ -162,10 +176,107 @@ function executeScriptInTestPage() {
...
+ * @property {Support} support
+ * @property {CommandsAPI_getATCommands} getATCommands
+ * @property {CommandsAPI_getModeInstructions} getModeInstructions
+ * @property {CommandsAPI_isKnownAT} isKnownAT
+ */
+
+/**
+ * @typedef BehaviorJSON
+ * @property {string} setup_script_description
+ * @property {string} setupTestPage
```
alflennik, line 250: This is the only property which isn't snake_case! Maybe there's a good reason for that but I can't tell!

tests/resources/aria-at-harness.mjs
```diff
@@ -176,266 +287,1179 @@ export function verifyATBehavior(atBehavior) {
...
+            `Restore default settings for ${at.name}. For help, read `,
+            {
+              href: "https://github.com/w3c/aria-at/wiki/Configuring-Screen-Readers-for-Testing",
+              description: "Configuring Screen Readers for Testing",
+            },
+            `.`,
+          ],
+          `Activate the "Open test page" button below, which opens the example to test in a new window${setupScriptDescription}`,
+        ],
+        boldInstructions: [modeInstructions, ...userInstructions],
```
alflennik, line 503: I think strongInstructions or emphasizedInstructions would be a better term given the fact this is typically used by AT users.

tests/resources/aria-at-harness.mjs
```diff
@@ -176,266 +287,1179 @@ export function verifyATBehavior(atBehavior) {
...
+ */
+
+/**
+ * @param {Behavior} testJson
+ * @param {Commands} commandJson
+ * @param {ResultState} resultState
+ * @param {(action: (state: ResultState) => ResultState) => void} dispatch
+ * @returns {InstructionDocument}
+ */
+export function instructionDocument(testJson, commandJson, resultState, dispatch) {
```
alflennik, line 457: The commandJson parameter isn't used!

alflennik (commented):
I'm really thankful you opened this up to early review Z, because this is going to take all my brainpower to review, haha, and definitely multiple passes. It's really amazing how you were able to turn this file upside down and right side up again. I think this file has been beyond mortal reach for a while, but with these changes I definitely feel like that is about to change.

tests/resources/vrender.mjs
s3ththompson: to echo @alflennik's comments, could you share your rationale for writing a custom implementation?

Any reason you didn't lean on a more commonly used dependency such as [lit-html](https://github.com/lit/lit/tree/main/packages/lit-html) or [snabbdom](https://github.com/snabbdom/snabbdom)?

tests/resources/aria-at-harness.mjs
s3ththompson: @alflennik could you share more about where command has been used to mean test instructions / assertions?

The working mode, harness, and current test format (e.g. `commands.json`) currently align on calling something like `SHIFT+TAB` a command (as in a "keyboard command").

alflennik: Command sequences would be an example: https://github.com/w3c/aria-at/issues/363 @s3ththompson

zcorpan: (1) makes sense to me. What needs to change to align with (1)?

alflennik: I'm going to resolve this. This is one of those comments where, by trying to explain my understanding, it helped me understand what I had been missing the whole time.

I think that number 1 is actually the meaning of command as we typically use it, and I realized there isn't really any second meaning floating around.

The second meaning was something floating around exclusively in my head, haha.

tests/resources/vrender.mjs
mzgoddard: > This file is pretty mind blowing! This ain't everyday JS, this is shiny stuff. I find myself wondering how you did this!

> I think, and correct me if I'm wrong, that this is a virtual DOM implementation to be used for rendering the tests, and I believe from our conversations that we wouldn't use it in the app, it would be confined to the aria-at repo.

Yes. aria-at-app will use the javascript object produced by `instructionDocument`. aria-at itself renders that same object with this virtual DOM implementation.

> Which of course, the thought of having to come in here to fix something is making me sweat slightly, haha. It's frankly over my head right now. So my question is, is there a way to avoid having to maintain this ourselves? I'm sure the reasoning is solid but I was hoping you would take a little time to help me grasp it.
> 
> Finally, I was wondering if there's a good way to test this file specifically. How do I put it through its paces so I feel like I gave it the review it deserves?



> to echo @alflennik's comments, could you share your rationale for writing a custom implementation?
> 
> Any reason you didn't lean on a more commonly used dependency such as [lit-html](https://github.com/lit/lit/tree/main/packages/lit-html) or [snabbdom](https://github.com/snabbdom/snabbdom)?

1. Maintain behavior
    - Focus on the page header on first render
    - On keyboard event on a checkbox focus an adjacent checkbox
2. Keep the probability of adding bugs in this change
    - Produce the same html structure
3. Minimal external dependencies
4. Easy to turn instructionDocument into the current html shape
    - Many adjacent elements represent a semantic section without a parent element
    - Fragments like in React give an easy way to produce a html structure with semantic and not structural sections
    - Fragments help break up long sequences of adjacent elements

- LitHTML
    - Adds one dependency
    - Cannot produce the same dom
        - Inserts html comments at dynamic content locations
            - I find it hard this would have an impact but conservatively I want to avoid it
        - Supporting focus behavior
            1. Option: Use components and their update hook
                - Components are web components (custom elements)
                - This would put a change into the html structure
            2. Option: Manually find the right element after rendering without components and focus it

- Snabbdom
  - Adds a minimum of one dependency (components are a separate dependency)
  - Does not support fragments https://github.com/snabbdom/snabbdom/issues/560
    1. Option: change the html structure, add wrapping elements
    3. Option: Use less helper functions and repeat many elements more manually

- vrender
  - Internal dependency
  - Fragment support

tests/resources/aria-at-harness.mjs
```diff
@@ -1,11 +1,12 @@
-import {commandsAPI} from './at-commands.mjs';
+import { commandsAPI } from "./at-commands.mjs";
+import { element, fragment, property, attribute, className, focus, render } from "./vrender.mjs";
 
 const UNDESIRABLES = [
```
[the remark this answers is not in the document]
mzgoddard: 👍 I think this works for this PR since it'll just change the internal references in this module.

tests/resources/aria-at-harness.mjs
```diff
@@ -162,10 +176,107 @@ function executeScriptInTestPage() {
...
+/**
+ * @typedef {{[mode in ATMode]: {[assitiveTech: string]: string;};}} CommandsAPI_ModeInstructions
+ */
+
+/**
+ * @typedef {([string] | [string, string])[]} CommandAT
+ */
+
+/**
+ * @typedef {{[atName: string]: CommandAT}} CommandMode
```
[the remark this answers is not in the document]
mzgoddard: :heart:

tests/resources/aria-at-harness.mjs
mzgoddard: `atName` sounds good.

tests/resources/aria-at-harness.mjs
mzgoddard: This type is the format of the test json files the test generation creates. So I think changing that key is off limits without a larger conversation.

tests/resources/aria-at-harness.mjs
mzgoddard: 👍

tests/resources/aria-at-harness.mjs
mzgoddard: 😅

tests/resources/aria-at-test-run.mjs
```diff
@@ -0,0 +1,1252 @@
...
+export const CommonResultMap = createEnumMap({
+  NOT_SET: "notSet",
+  PASS: "pass",
+});
+
+/**
+ * @typedef {EnumValues<typeof AdditionalAssertionResultMap>} AdditionalAssertionResult
+ */
+
+export const AdditionalAssertionResultMap = createEnumMap({
```
howard-e, line 457: nit: I certainly agree with the separation of concerns for the `Map`s here but it does feel like a few of these could be packaged into 1, especially the `...AssertionMap`s because they are related and it could limit the amount of documentation/'looking for property in wrong map' errors one could possibly encounter? Just a thought.

tests/resources/aria-at-test-run.mjs
```diff
@@ -0,0 +1,1252 @@
...
+  ...CommonResultMap,
+  FAIL_SUPPORT: "failSupport",
```
howard-e, line 459: nit: I noticed the values of these generated `Map`s all seem to differ between camelCase, natural language and capitalized. I figure it has to do with where their respective values will be used. Perhaps their appropriate usage could be described in their JSDoc tags?

tests/resources/vrender.mjs
```diff
@@ -0,0 +1,980 @@
...
+    }
+  ),
+  softTeardown: /** @type {SoftTeardownFunction} */ (softTeardownElement),
+};
+
+/** @type {FragmentStateType} */
+const FragmentType = {
+  name: FRAGMENT_TYPE_NAME,
+  diffEntry: /** @type {DiffEntryFunction} */ (diffChildEntry),
+  init(queue, parent, after, node) {
```
howard-e, line 246: The `node` param isn't being used here.

howard-e (commented):
Outside of nitpicks and what's been spoken about before, I'm finding it difficult to add much from a technical standpoint. This is without a doubt, impressive! I also believe it has been tested well enough to some degree to be convinced it does exactly what it has been proposed to do.

However, my general concern is around the maintainability of it which to also echo @alflennik, should any additional work ever be required. I know things are still changing but maybe some additional descriptions should start getting added in with the newly added `JSDoc` tags (which is impressive in and of itself as well) to describe further what these inputs are, especially the `map`, `object` and `any` properties. This should be helpful to others down the line in helping to alleviate this concern.

In a previous meeting we had as well, you provided very useful descriptions on the overall purpose of each file which I believe helped my understanding immensely going in. So having a similar write-up could be helpful to capture your initial thoughts for others as well I believe.

tests/resources/aria-at-test-run.mjs
mzgoddard: Yeah. The natural language, spaced and title cased, values and the all capitalized values are values as they already appear in the json object built and submitted from the test result from input. I'll add something like that to the doc tags.

tests/resources/aria-at-test-run.mjs
mzgoddard: Hmm. Yeah at this point I just wanted to mirror the values that get exported with ones in the internal state that follow a common idiom.

PR #442, opened 2021-06-04
Add script to count test plans, tests, assertions, and command/assertion pairs

s3ththompson:
[Preview Tests](https://deploy-preview-442--aria-at.netlify.app)

[Preview Tests](https://raw.githack.com/w3c/aria-at/add-count-script/index.html)

This PR adds a script to the scripts directory which provides basic metrics for all test plans or a single specified test plan.

The output of this script can be used to track the growth of the ARIA-AT repository over time.

Example output:
```
$ npm run count
8 test plans, 242 tests, 919 assertions, 1643 command/assertion pairs
```


jscholes:
> 1643 command/assertions

@s3ththompson What does that mean?  What is a "command/assertion"?

s3ththompson:
@jscholes was trying to think of a better name for that... it's basically number of commands times number of assertions in order to count the total number of distinct data points that are collected. i'll happy replace the wording if you have any suggestions!

jscholes:
@s3ththompson Hmm, tricky one.  Reading your response, what comes to mind immediately is "command/assertion pairs".  Or just "datapoints", which is so vague that it all but forces users to read the documentation (and would need to be documented accordingly).

s3ththompson:
@jscholes I like "command/assertion pairs" better. I updated the script to use that terminology.

scripts/count-assertions.js
```diff
@@ -0,0 +1,133 @@
...
+    .readdirSync(testsDirectory)
+    .filter(f => f !== 'resources' && fs.statSync(path.join(testsDirectory, f)).isDirectory());
+
+  for (let directory of directories) {
+    promises.push(count(path.join(testsDirectory, directory)));
+  }
+}
+
+// wait for all promises to complete, then read totals
+Promise.all(promises)
```
mzgoddard, line 123: I think instead of `promises` we should store the `directories`. For the single directory that would be a single entry array like `[directory]`.

Something like:

```js
let directories;

if (args._.length) {
  // ...

  directories = [directory];

  // else, count all directories
} else {
  directories = fs
    .readdirSync(testsDirectory)
    .filter(f => f !== 'resources' && fs.statSync(path.join(testsDirectory, f)).isDirectory());
}

Promise.all(directories.map(name => count(path.join(testsDirectory, name))))
```

scripts/count-assertions.js
```diff
@@ -0,0 +1,133 @@
...
+            reject(error);
+          });
+      })
+      .catch(error => {
+        reject(error);
+      });
+  });
+}
+
+let promises = [];
```
mzgoddard, line 96: What do you think about wrapping the remaining body of this script in an `async` function?

mzgoddard (changes requested):
I think some changes would be good.

mzgoddard:
I am going to post a commit here with the suggested changes.

scripts/count-assertions.js
```diff
@@ -0,0 +1,133 @@
...
+
+    let numCommands = 0;
+    for (let letter of ['A', 'B', 'C', 'D', 'E', 'F']) {
+      if (command[`command${letter}`]) {
+        numCommands++;
+      }
+    }
+
+    let numAssertions = 0;
+    for (let j = 0; j < 7; j++) {
```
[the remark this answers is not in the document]
mzgoddard: Its 30 assertions.

scripts/count-assertions.js
```diff
@@ -0,0 +1,133 @@
...
+`);
+  process.exit();
+}
+
+const testsDirectory = path.resolve(__dirname, '..', 'tests');
+
+let totalTestPlans = 0;
+let totalTests = 0;
+let totalAssertions = 0;
+let totalCommandAssertions = 0;
```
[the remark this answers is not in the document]
mzgoddard: I'll file an issue for something on this.

mzgoddard (approved):
I've pushed a commit that applies the suggestions I previously made, along with some other changes.

Looking forward to your thoughts @howard-e.

scripts/count-assertions.js
```diff
@@ -0,0 +1,163 @@
...
+function countTests(tests, commands) {
+  let totalTests = 0;
+  let totalAssertions = 0;
+  let totalCommandAssertions = 0;
+
+  for (const [i, test] of tests.entries()) {
+    const command = commands[i];
+
+    let numCommands = 0;
+    for (const letter of ['A', 'B', 'C', 'D', 'E', 'F']) {
```
howard-e, line 52: Since merging #832, `commandG - commandZ` (up to 26 commands) are supported and at least 2 tests at this time support up to 7 commands (`commandG`).

Will push a commit to update this.

howard-e (commented):
Thanks @mzgoddard for the discussion and additional changes you provided here!

howard-e (approved):

scripts/count-assertions.js
howard-e: See https://github.com/w3c/aria-at/pull/442/commits/aaaeb6f7714fc6ba276f6c1c397ffb4f4769afdb.

Issue #422, opened 2021-04-08
Backslash doesn't work as separator in references.csv


jscholes:
When using the `update-reference.js` script, it prints out instructions advising how `references.csv` should be updated to point to the new timestamped example file:

```
console.log(`Reference file created at tests/${exampleName}/${referenceHtmlPath}.\nTo switch the test to run the updated reference:\n\t1. Commit this change\n\t2. Update ${path.join('tests', exampleName, 'data', 'reference.csv')} with the reference ${referenceHtmlPath}\n...`);
```

Where `referenceHtmlPath` is the result of a call to `path.join`, using `path.sep` (the platform-specific separator string).  On Windows, this causes the output to contain a path separated by backslashes.  But when this is pasted into `references.csv`, it won't work because the tests are previewed in an environment powered by Linux (which is also where I believe the test runnre is executed).  When attempting to use the PR preview feature, for instance, loading any of the test pages results in an HTTP/404 error.

howard-e:
A follow-up description on this issue is that cases may arise where a "backslash path separator" may be a part of the `reference` field value in `references.csv`. When processing these values, we should ensure that the path is transformed to a 'POSIX-based' standard.

s3ththompson:
@jscholes did you have further thoughts about @mcking65's suggestion that we standardize around paths in `references.csv` needing to be formatted as relative URIs (e.g. browser-friendly relative paths), rather than filepaths? The output of the test import script should always print relative URIs after #446 is merged.

As for the workflow issue of not being able to copy/paste from the Windows file explorer, it seems like it would be a better solution to have some sort of locally run conversion script / tool rather than committing references.csv files to the repo with differing filepath conventions (some Windows, some URI, some POSIX, etc.). Would something like this be an acceptable workaround?

jscholes:
@s3ththompson It is easier to manually convert a Windows path to a POSIX path by hand, than trying to manually create a file URI.  I.e. if a member of the PAC team is using Windows and wishes to manually create/update references.csv, changing some backslashes to forward slashes in a path they've copied to the clipboard is trivial.  Converting that copied path to a file URI is not.

Now, a reasonable response to that may be, "always use the update-reference script", and then we can use whatever format we want no matter how human-friendly it is or is not.  But I'm conscious of the fact that the script only supports the APG for now, rather than, say, custom import locations for smaller custom examples.  So updating the references file by hand may be necessary for a while.

[...]

PR #446, opened 2021-06-10
Fix for backslash separator in update-reference script output

howard-e:
[Preview Tests](https://raw.githack.com/w3c/aria-at/fix-422/build/index.html)


jscholes:
@howard-e Thanks for working on this, but it doesn't really fix #422 exactly.  If somebody manually copies a Windows path into `references.csv`, which I do almost all the time, it still has to be modified because `references.csv` only accepts forward slashes.  This only fixes the case where someone is using the script output, which won't be 100% of the time.

Is it possible to make `references.csv` platform-agnostic in terms of separators?  This will reduce likely user error quite a bit.

howard-e:
Thanks for the feedback @jscholes. I wasn't aware that was a case to be accounted for as #422 wasn't covering that particular scenario. But perhaps we could track that platform-agnostic approach for the processing of `references.csv` in a separate issue? I believe it may be significant enough for such.

jscholes:
@howard-e The issue title is "Backslash doesn't work as separator in references.csv", although admittedly the body does talk more about the script.  So if we do go for a different issue, #422 will need to be renamed.

howard-e:
Good point. How about we keep the issue open and expound on the description there?

jesdaigle (approved):

s3ththompson:
@jscholes i renamed #422 (which will remain open after this PR is merged). would you mind reviewing this PR just as it applies to the normalizing the formatting of stdout for paths?

jscholes (approved):

jscholes:
@s3ththompson Thanks, and approved.  I didn't realise this was blocking on me, apologies for that.

s3ththompson:
@jscholes no problem, we hadn't explicitly added you as a reviewer until today, so it wasn't blocking on you. we'll make sure to explicitly request reviews going forward.

Issue #443, opened 2021-06-09
Review of existing tests' ability to be automated


zcorpan:
## Background

In the [ARIA-AT Automation Design Doc](https://docs.google.com/document/d/1YxHAf6r3E2RG4REaYcmM14DuxfBgP4XoWRdfSk5aAQM/edit#heading=h.23ije5xdfmzx), we discussed that the existing ARIA AT tests may need some changes to allow for automation without ambiguity in the test instructions. This issue reviews the existing tests and suggests changes we think are necessary for automation. These changes could make the tests clearer for manual testers as well.

View these suggestions as a starting point for discussion; they are not set in stone.

## Multiple tests

*   Between each key to press, there is currently an implied “reset”, which should be explicit. With a reset button as suggested in [issue #358](https://github.com/w3c/aria-at/issues/358), this could be solved. The instruction can be to activate the reset button between each key in the list of keys to press.
*   For JAWS:
    *   “Verify the Virtual Cursor is active by pressing Alt+Delete. If it is not, turn on the Virtual Cursor by pressing Insert+Z.”
        *   For automation, the environment ought to be known. If Virtual Cursor is active by default, the test doesn’t need to do anything. If it’s not, always press Insert+Z.
*   For NVDA:
    *   “Insure NVDA is in browse mode by pressing Escape. Note: This command has no effect if NVDA is already in browse mode.”
        *   Typo: Insure should be “ensure” or “verify”.
        *   For automation, the environment ought to be known. If browse mode is active by default, the test doesn’t need to do anything. If it’s not, always press Escape.

## checkbox

*   Should split “navigate forwards” and “navigate backwards” to separate tests.
*   For tests that navigate forwards, insert a link before the checkbox and set focus to it with the setup script. Similar to checkbox-tri-state.
*   For tests that navigate backwards, insert a link after the checkbox and set focus to it with the setup script. Similar to checkbox-tri-state.
*   Remove “Note: it should be in the unchecked state.” from the instructions. This is covered in the success criteria.

## checkbox-tri-state

*   Remove “Note: it should be in the partially checked state.” from the instructions. This is covered in the success criteria.
*   Remove “You may need to press certain commands (such as Down Arrow) multiple times to hear all information relating to the checkbox.” and change the key instructions with exactly how many key presses for each key is expected. With [PR #438](https://github.com/w3c/aria-at/pull/438) this is possible without adding new keys to tests/resources/keys.mjs.

## combobox-autocomplete-both

[...]

Issue #394, opened 2021-03-10
Document APG example modification process


jscholes:
When examples are pulled into the ARIA-AT repo from [aria-practices](https://github.com/w3c/aria-practices), we currently modify them as follows:

1. Remove core JS/CSS references from the head section of the example HTML file.  This avoids errors with trying to pull in CSS/JS which is common to all APG examples but doesn't exist in the ARIA-AT repo, makes sure we don't get the CodePen button, etc.
2. Remove any CSS, JS and image asset files pulled in by the [update-reference script](https://github.com/w3c/aria-at/blob/master/scripts/update-reference.js) which the example doesn't actually need.
3. Remove the "Related Links" nav and similar example links.
4. Remove any unnecessary text from the preamble.  Usually, this involves leaving the first paragraph describing the pattern intact, but getting rid of everything else.
5. Update the `href` of the `<a>` linking to the design pattern in the APG, to reference `https://w3c.github.com/aria-practices`, rather than a relative, repo-specific path like `../../#something`).
6. Remove the "Accessibility Features", "Keyboard Support", "Role, Property, State, and Tabindex Attributes" and source code sections.
7. Remove the nav at the bottom of the example (after the main).
8. Update the page title to remove the reference to the WAI-ARIA version, because the examples in the ARIA-AT repo are pulled in from the main APG branch and versioned via a timestamped directory instead.
9. Add links before and after the component after test, to be used in setup scripts.

The above is quite a bit of work, but can also be complicated by different examples following different structures or relying on assets that others do not.  For example, modifying the [Data Grid Examples](https://w3c.github.io/aria-practices/examples/grid/dataGrids.html) for PR #393 involved more work, because:

1. The code powering the grids seems to rely on shared functionality from [utils.js](https://github.com/w3c/aria-practices/blob/main/examples/js/utils.js).  As this is a file shared by multiple APG examples and therefore not a child of the `grid` directory, the `update-reference` script didn't pull it in to begin with.
2. The example page has multiple data grids, some of which have more functionality than the first relatively simple one that was actually under test.  This required more care and testing to determine which assets were needed by all grids or only some of them, e.g. the CSS contained references to image files that were only needed by the second and third grids.
3. The example could only be successfully tested while running a local HTTP server, because some of the bundled assets used functionality that is disallowed by a browser when executing JS from a local file.

Finally, it should be noted that the update-reference script does print some instructions for test writers, as follows:

[...]

PR #447, opened 2021-06-10
Create updated tests for APG design pattern example: checkbox

zcorpan:
Part of #443

---

This is currently a work-in-progress. The changes in `commands.csv` at least are not complete.


zcorpan:
For some reason `npm run create-all-tests` doesn't create the tests for `checkbox`. I've probably made a mistake somewhere, but there's no useful error message. I haven't yet figured out what's wrong.

zcorpan:
OK, I think everything works now! Ready for review.

jscholes:
@zcorpan 

#### Example

1. The `reference` directory contains assets relating to the tri-state checkbox (checkbox2) as well as the one under test.
2. The default JS/CSS references haven't been removed from the example (e.g. the W3C stylesheet, references to JS assets from the APG repo that we don't use).
3. The example has been modified to the degree that the heading structure now skips levels (`<h1>` followed by `<h3>`).  Usually we don't cut this much content out.

For more info, see https://github.com/w3c/aria-at/issues/394.  Some of these issues seem to have already been present in the checkbox plan, but we should probably get them fixed up if we're submitting updates to it.

#### commands.csv

The JAWS commands still include smart navigation commands which are no longer in scope.

#### tests.csv

1. The navigation tests don't include any state in their names.  E.g. test 1 and test 9 have the same title ("Navigate forwards to checkbox in reading mode").
2. The inclusion of an article in test titles is inconsistent.  E.g. "Navigate forwards to checkbox" versus "Operate a checkbox" (no "a" in the first one).
3. The "operate a checkbox" tests aren't separated based on the state transition.  I.e. we need tests to check a checkbox, and separate ones to uncheck it.

I have some additional thoughts about the group-related tests, but I'll revisit those.

#### Test Preview

The index page is working, but the [review page for checkbox](https://raw.githack.com/w3c/aria-at/bocoup/update-checkbox-tests/build/review/checkbox.html) seems completely broken.  For example I see:

* Test 1: Navigate forwards to checkbox in reading mode
* Test 2: Navigate to an unchecked checkbox in reading mode

The second of those isn't even present in `tests.csv`?  So I have no idea where it's coming from.  The same applies all the way down, with old tests interleaved with the new.

zcorpan:
I've addressed the comments on the source files.

I don't know what was up with the test preview.

@jscholes 

> I have some additional thoughts about the group-related tests, but I'll revisit those.

What about them? :)

Issue #358, opened 2020-12-16
How to reset state within a test


zcorpan:
In  https://github.com/w3c/aria-at/issues/349#issuecomment-737902356 I asked:

> Is the list of commands above match how we think a tester ought to run the test (when running manually)? I've made the assumption that a tester would want to "reset" between different key commands to perform the task, and used "h" and "Shift+h" for that to navigate to a heading.

This applies both when testing manually and in an automated setup.

@jscholes replied that it may be better to reload the test.

I think we haven't come to a conclusion on this, so opening a dedicated issue.

jscholes:
@zcorpan Thanks for opening this.  I think expecting human testers to infer the right way to go about testing multiple commands is fair when we're dealing with navigation to a specific element, or with simpler tests.  My concern is that as the patterns under test become more complex, the way testers will have to reset becomes more challenging.  For example:

* We ask testers to transition a control from state A to state B using two commands, Space and Enter.  It is implied that once the control is in state B after the first command, the tester will have to place it back in state A to carry out the second.  This is not as obvious as just finding a heading, and may cause testers to inadvertently test the transition from state B to state A with the second command.
* We use a setup script to carry out several steps, such as expanding a combobox and placing focus on a specific item within the popup, or opening multiple nested modals.  We do include a description of the setup script's actions, but it's not guaranteed that testers will follow them correctly or at all.  In essence, a setup script crafted to increase consistency will only be executed for the first command in some cases.

I know there is some school of thought that setup script usage should be limited, and that we should trust human testers to put up some of the scaffolding themselves.  But for automation, those steps will need to be scripted regardless, so I'm in favour of setup scripts assisting human testers to the greatest feasible degree.

There are multiple options that come to mind:

1. Make each command its own test, so that users will be forced to open the test page from scratch each time.  This would be a usability nightmare for human testers considering that some tests have up to six commands.
2. Ask testers to reload the page between commands.  At present, I don't believe setup scripts to be executed when just pressing F5 or similar, so that would need to be resolved if that is the case.
3. Prompt testers to explicitly close the test page between each command, perhaps by reorganising test pages to include the button to open the page multiple times rather than just once.
4. Give human testers explicit instructions about how to reset state between commands.  I'm not a fan of this, as we'll either need to:
    * provide two set of these instructions, one for automation and one for humans; or
    * come up with an abstracted format which is machine-readable for automation but can also output human-readable instructions for testers.

[...]

Issue #369, opened 2020-12-18
Potential setup script improvements


jscholes:
On a recent ARIA-AT CG call, we were discussing how page state should be reset between tests (see #358).  Tied into that discussion were some thoughts about possible improvements to setup scripts, both in terms of how they are written and executed:

* We need some method of sharing logic across setup scripts or chaining them for a test?  For instance, the updated test plan for the editable combobox example (https://github.com/w3c/aria-at/pull/355) includes 11 separate scripts which set focus on the combobox, and 8 which also expand the control.  It would be great to write shared logic once, and indicate that it applies to a test in addition to any custom code for that test.  Sequencing scripts may be the simplest approach.
* Do we need a "tear down"/"reset to known good state" mechanism?  In the related issue about resetting page state, the most recent proposed solution that multiple people agree with is a button on test pages to run the setup scripts.  But there is a risk that any steps carried out by a previous script invocation won't be sufficiently undone, such as hiding an element or changing the accessible name of something.  Note that we don't currently do the latter in any tests, but we may in the future.

robfentress:
> But there is a risk that any steps carried out by a previous script invocation won't be sufficiently undone, such as hiding an element or changing the accessible name of something.  Note that we don't currently do the latter in any tests, but we may in the future.

Do you have an example of where this may be the case or is this a more hypothetical concern?

jscholes:
@robfentress 

> Do you have an example of where this may be the case

Changes are made to pages by setup scripts in a number of our test plans to date.  E.g.:

* In select-only or editable combobox, the combobox value is set to something other than the default.
* For checkboxes and similar controls, the state is set on page load and users are then prompted to change it.  But tests are targeting a specific state transition like checked to not checked, so the state must be reset.
* In disclosure FAQ, one of the setup scripts purposefully hides the "Navigate backwards from here" link so it doesn't get in the way during a test for which it isn't relevant.

jscholes:
Actionable next step for myself: write up an APG issue explaining current problem and suggested route forward.  Namely, a global object for interacting with APG components.

mcking65:
Another thing I love about this approach is that the "shortcut" for the reset button is simply use the browser refresh key. So, the process is explicitly defined on the page, but the experienced tester can easily use the shortcut if they prefer.

mzgoddard:
[...]

PR #450, opened 2021-06-15
add Setup Test Page button into test window

mzgoddard:
[Open PR Preview](https://raw.githack.com/mzgoddard/aria-at/mzgoddard-reset-test/index.html)

This change injects a button at the bottom of the test page in the window opened by activating the `Open Test Page` button.

The button has the text `Setup Test Page` and the autofocus attribute. Clicking it will run a test setup script if this test has one and then disables itself. If there is no setup script the button will disabled at the start.

<details>
<summary>Original PR Description</summary>

This change is a first step for https://github.com/w3c/aria-at/issues/358. In a follow up PR, we can inject a button into the test page that on click will reload the test page window.

After the test page window is reloaded, or refreshed, the setupTestPage script will be executed on the reloaded page. As well the unload listener is added again so that a second reload, and third and so on, will have the same functionality. By adding the unload listener after reloading it will also be able to re-enable the open test page when it is closed.
</details>

### DIFF
--- a/scripts/review-template.mustache
+++ b/scripts/review-template.mustache
@@ -18,8 +18,10 @@
       }
 
     </style>
-    <script>
-      const filterATs = function(button) {
+    <script type="module">
+      import {TestWindow} from '../tests/resources/aria-at-test-window.mjs';
+
+      window.filterATs = function(button) {
         const val = button.value;
 
         if (val === 'all') {
@@ -54,45 +56,30 @@
 
           }
         }
-      }
+      };
 
-      function executeScriptInTestPage(setupScriptName, testPageWindow) {
-        if (testPageWindow.location.origin !== window.location.origin // make sure the origin is the same, and prevent this from firing on an 'about' page
-            || testPageWindow.document.readyState !== 'complete'
-        ) {
-          window.setTimeout(() => {
-            executeScriptInTestPage(setupScriptName, testPageWindow);
-          }, 100);
-          return;
-        }
+      var scripts = {
+        {{{setupScripts}}}
+      };
 
-        if (setupScriptName) {
-          var scripts = {
-            {{{setupScripts}}}
-          };
+      window.openTestPage = function(testPageUri, testNumber, setupScriptName) {
+        var openTestPageElement = document.getElementById('open-test-page-' + testNumber);
 
-          scripts[setupScriptName](testPageWindow.document);
-        }
-      }
-
-      function openTestPage(testPageUri, testNumber, setupScriptName) {
-        testPageWindow = window.open(testPageUri, '_blank', 'toolbar=0,location=0,menubar=0,width=400,height=400');
-
-        document.getElementById('open-test-page-' + testNumber).disabled = true;
-
-        // If the window is closed, re-enable open popup button
-        testPageWindow.onunload = function(event) {
-          window.setTimeout(() => {
-            if (testPageWindow.closed) {
-              testPageWindow = undefined;
-              document.getElementById('open-test-page-' + testNumber).disabled = false;
+        var testWindow = new TestWindow({
+          pageUri: testPageUri,
+          setupScriptName,
+          scripts,
+          hooks: {
+            windowOpened() {
+              openTestPageElement.disabled = true;
+            },
+            windowClosed() {
+              openTestPageElement.disabled = false;
             }
-          }, 100);
-
-        };
-
-        executeScriptInTestPage(setupScriptName, testPageWindow);
-      }
+          }
+        });
+        testWindow.open();
+      };
     </script>
   </head>
   <body>
--- a/tests/resources/aria-at-test-window.mjs
+++ b/tests/resources/aria-at-test-window.mjs
@@ -28,46 +28,50 @@ export class TestWindow {
     this.scripts = scripts;
   }
 
-  /** If the window is closed, re-enable open popup button. */
   windowOnBeforeUnload() {
     window.setTimeout(() => {
       if (this.window.closed) {
         this.window = undefined;
-
-        this.hooks.windowClosed();
-      } else {
-        // If the window is open (after a location.reload()) rerun the setupTestPage script.
-        this.prepare();
       }
+
+      /** Re-enable open popup button. If the window was reloaded it may be 100ms  */
+      this.hooks.windowClosed();
     }, 100);
   }
 
+  windowOnDOMContentLoaded() {
+    if (this.window) {
+      this.prepare();
+    }
+  }
+
+  /**
+   * @param {Window} testPageWindow
+   */
   injectSetupResetButton(testPageWindow) {
+    const {scripts, setupScriptName} = this;
+    const setupScriptPresent = Boolean(setupScriptName && scripts[setupScriptName]);
+
+    /** @type {'setup' | 'post-setup'} */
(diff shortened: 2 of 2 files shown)


jscholes:
@mzgoddard Sorry, but this isn't the approach the community group reached consensus on.  #358 is outdated, and #369 is the more appropriate reference.  In summary:

1. Setup scripts should never be executed on page load.  This causes screen reader issues, script flakiness, among other things.
2. There should be a button on the test page that runs the setup script, which testers will need to manually press.
3. Once pressed, the button can transition to being a "Reset" button, which only refreshes the test page and allows the execute button to be pressed again.  If a tester chooses to manually refresh the page instead to restore the execute button, that's also fine.

s3ththompson:
@jscholes @mzgoddard I think Z plans to add the actual button (with the setup / reset behavior) in a followup PR. We discussed the consensus from #369 and again as summarized in your issue. I think perhaps the point of confusion was: how can we prevent the script that dynamically injects the setup script button from being flaky itself? Do you have any context on how to safely inject a new button element into the example?

jscholes:
@s3ththompson Unless the wording of the initial comment on this PR is inaccurate, I don't agree with your assessment.  Specifically:

> After the test page window is reloaded, or refreshed, the setupTestPage script will be executed on the reloaded page.

... and ...

> In a follow up PR, we can inject a button into the test page that on click will reload the test page window.

What I'm reading here is:

1. this PR allows the setup script to be executed on refresh, and
2. a follow-up PR will add an explicit refresh button, which relies on the behaviour introduced in this one, and
3. testers will only need to press that explicit button once, because it will reload the page and exec the setup script in one go.

If my understanding is wrong, please let me know and update the PR description.  Otherwise, if I'm reading the situation correctly, I maintain that this isn't the approach we want to proceed with.

tests/resources/aria-at-harness.mjs
```diff
@@ -161,10 +162,38 @@ function executeScriptInTestPage() {
...
+  const buttonDiv = testPageWindow.document.createElement('div');
+  buttonDiv.innerHTML = `
+    <div style="position: relative; left: 0; right: 0; height: 2rem;">
+      <button style="height: 100%; width: 100%;">Run Test Setup</button>
+    </div>
+  `;
+  const button = buttonDiv.querySelector('button');
+
+  /** @type {'setup' | 'reload'} */
+  let runSetupOrRelad = 'setup';
```
jscholes, line 174: You have a small typo here, "Relad" -> "Reload".

jscholes:
@mzgoddard I notice that there is no preview for this PR.  What's going on there?

mzgoddard:
> @mzgoddard I notice that there is no preview for this PR. What's going on there?

I think that was this PR not having rebased onto the #441 revert (by #449). Unless you are referring to something else.

jscholes:
That definitely sounds like the likely cause.  Just noting that previewing this functionality will be critical.

mzgoddard:
@jscholes I updated the PR's description. I updated this change to inject a button to the top of the test page that on a first click runs the setup script if there is one, and on a second click reloads the page.

I suspect this still has some of the execution flakkiness of the current setup script execution. I've seen in my testing on Firefox, on Mac, that on test page reload the button may not inject. I suspect this has to do with the way the setup script was automatically run and now how the button is injected. I haven't seen this in Chrome, but that may be due to a smaller sample size as I'm mostly testing in Firefox.

And I think it may cause some issue with screen readers with the button being injected.

If these issues do persist in this change we probably have to change how the button is inserted, primarily when it is inserted. The easiest option may be to modify the reference pages that tests are performed in, but I am trying to avoid that.

jscholes:
@mzgoddard Maybe I'm missing something?  I looked at some tests from the Modal Dialog plan, both via the Index and Review pages.  But the setup scripts are still automatically executed when I open the test page, and there are no new buttons anywhere.

richnoah:
@jscholes would you please convey what URL you are using to view the tests for the Modal Dialog Plan to aid in diagnosing the issue?

jscholes:
@richnoah The preview page at: https://raw.githack.com/mzgoddard/aria-at/mzgoddard-reset-test/review/modal-dialog.html

mzgoddard:
@jscholes Ah. This should have been obvious to me. I had been confirming the change on test pages and not on review pages.

The review page doesn't use the aria-at-harness module to open the test page window. So it currently needs its own change to the script it uses.

jscholes:
@mzgoddard Makes sense.  Is it planned for the review pages to use the harness to cut down on duplication and other corner cases in the future?

In terms of the implementation of the button when using test pages:

1. The button needs an `autofocus` attribute to ensure that it receives focus automatically at page load.  I stress that it should use `autofocus`, rather than `element.focus`, as it's then up to the browser to ensure that it works correctly.
2. The button isn't within the main landmark region, which will cause extra speech verbosity when activating it moves focus onto something within main.  Can it be injected into the `<main>` container?

Other than the above, it seems to be working great.

s3ththompson:
> The button isn't within the main landmark region, which will cause extra speech verbosity when activating it moves focus onto something within main. Can it be injected into the <main> container?

@jscholes It would be great if this script was agnostic to the format of the example (to prepare for examples beyond APG...). Will all examples have `main` elements? Or should the button injection code try to find a `main` element, but fall back to the `body`?

> I've seen in my testing on Firefox, on Mac, that on test page reload the button may not inject.

@mzgoddard I am able to reproduce this issue consistently on Chrome. The 100ms timeout seems like it might be an issue. Can the parent asynchronously wait until the window has been loaded (or reloaded) before triggering (or retriggering) the script?

jscholes:
> should the button injection code try to find a main element, but fall back to the body?

This sounds like a reasonable approach.

> Can the parent asynchronously wait until the window has been loaded (or reloaded) before triggering (or retriggering) the script?

I'm concerned about the use of `autofocus`, which is important to the intended design, if the button isn't present at page load.  Is it not possible to somehow construct the mark-up before presenting it to be rendered, rather than "injecting" the button into an existing document?

mzgoddard:
Was able to revisit this and made some changes. A major change is removing the attempt to support reloading the window. More on that after a list of other differences.

Changes pushed to this PR:
- add button in `main` element if it exists, otherwise add to`body` element
- add button on DOMContentLoaded event, when the dom can first be modified
- use TestRunWindow in review pages, using the same code to open the window and add setup button
- disable setup button if there is no setup script or once it is activated
- add autofocus attribute to the setup button

I think after visiting this a few times, before we can use a technical solution to support "resetting" the test page we probably need a change to the process of making test plans. I think there are two non-exclusive changes that could be made to the process of making test plans. I'll write about them in #369 as we can treat it as a separate set of work.

jscholes:
@mzgoddard 

> I think after visiting this a few times, before we can use a technical solution to support "resetting" the test page we probably need a change to the process of making test plans.

I'm intrigued, so I'll eagerly await updates on #369.  Right now, it seems like one of the simple parts of the web, i.e. refreshing a page, doesn't work because of how the windows are being opened, kind of as pop-ups rather than straight pages.  Is there really no way we can just have the page open by URL, and include some JS files that can handle setup script duties?  That would also make testing them a lot easier, because it would mean not having to go through the testing flow to get there.

PR #463, opened 2021-07-22
Remove non-semantic visual code formatting

jugglinmike:
A number of text files in this repository use empty space characters in
places where they have no technical significance. This practice is
generally followed to vertically align information when text is rendered
visually. For users of assistive technologies, it may not be apparent
that this practice is being followed, and it may be burdensome to
maintain the convention.

Remove the formatting to promote equity in the project's contribution
experience.

---

`git diff` and `git show` accept a flag `-w` which hides changes that only impact empty space. That's a useful feature to review patches like this because it helps demonstrate that there aren't any unintended changes amidst all the reformatting.


jscholes:
@jugglinmike Thanks for taking care of this.  I would also ask, along the same lines, that we don't use hard line wrapping in PR or issue comments either, which exhibits the same problem.  I.e. my screen reader can't present the text with the configured line length if it includes hard returns at a certain number of columns.  Thanks 👍

jugglinmike:
Thanks for letting me know! I'll take care to avoid hard line wrapping when posting to GitHub.

PR #483, opened 2021-09-20
Export TestResult json with TestRunInputOutput

mzgoddard:
[Preview Tests](https://raw.githack.com/w3c/aria-at/mzgoddard-test-result-json/build/index.html)

Export and import a json format of a test result without a test's verbose instruction information. This format is intended to make integration with aria-at-app easier.

This change includes:
- Improve JSDoc compatible typedef for AriaATTestRun.State (was TestRunState) in `tests/resources/types/aria-at-test-run.js` following typedefs/namespaces in `types`.
- Add JSDoc namespace `AriaATTestResult` in `tests/resources/types/aria-at-test-result.js` containing `AriaATTestResult.JSON` the import and export ttype for the new TestRunInputOutput methods.
- Forward errors from instruction document to test page document.
- Add `resultFormat` and `resultJSON` ConfigInput methods, that aria-at-harness sets with query params, as an avenue to test AriaATTestResult.JSON import and export. `resultFormat` isn't required by the new TestRunInputOutput methods.

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -40,7 +40,7 @@
             <td><a href="./tests/checkbox/index.html">Index</a></td>
             <td><a href="./review/checkbox.html">Review</a></td>
             <td>26</td>
-            <td><a href="https://github.com/w3c/aria-at/commit/2119af6" target="_blank">2119af6 Merge remote-tracking branch &#39;origin&#x2F;issue-471&#39; into issue-471
+            <td><a href="https://github.com/w3c/aria-at/commit/5eddca5" target="_blank">5eddca5 fixup! add TestRunInput methods for TestResultJSON data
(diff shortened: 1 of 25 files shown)


alflennik (approved):
I had attempted to implement remapping something like this around a week and a half ago, and I'm much happier with what you came up with here than my previous attempts.

I was concerned that the harness systems might need more information than the TestResult type could provide, but it seems like you were able to skillfully sidestep any problems there.

The big test of this new functionality will come when we attach this new result type to the App's Test Queue pages, but I feel like that shouldn't get in the way of merging this into ARIA-AT.

howard-e (approved):
Looks good!

PR #485, opened 2021-09-21
 add Setup Test Page button into test window

mzgoddard:
[Preview Tests](https://raw.githack.com/w3c/aria-at/mzgoddard-reset-test/build/index.html)

Close #450

This is same work in PR #450. This PR replaces that one so that our github workflow can run as expected.

> This change injects a button at the bottom of the test page in the window opened by activating the `Open Test Page` button.
>
> The button has the text `Setup Test Page` and the autofocus attribute. Clicking it will run a test setup script if this test has one and then disables itself. If there is no setup script the button will disabled at the start.

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -40,7 +40,7 @@
             <td><a href="./tests/checkbox/index.html">Index</a></td>
             <td><a href="./review/checkbox.html">Review</a></td>
             <td>26</td>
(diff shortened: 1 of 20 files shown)


jscholes:
@mzgoddard Just so I'm clear: does this PR introduce the short-term fix, of still injecting the button into test case pages from the test runner?  If so, has the more long-term fix, of working this into the build process, been added to the roadmap for later?

PR #493, opened 2021-09-22
Exclude 'rootDirectory' path from generated files

zcorpan:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/remove-root-directory-path/build/index.html)

This path causes conflicts and exposes unnecessary information from contributors.

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -40,7 +40,7 @@
             <td><a href="./tests/checkbox/index.html">Index</a></td>
             <td><a href="./review/checkbox.html">Review</a></td>
             <td>26</td>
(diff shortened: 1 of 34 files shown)


mzgoddard (approved):
Not sure if it should be a part of this change. But there is no `start` script in this package currently.

howard-e (approved):

zcorpan:
@mzgoddard can you file a new issue about that?

zcorpan:
I've removed the mention of `npm run start` also

mzgoddard (approved):

PR #512, opened 2021-09-28
Render reference pages with setup script and button

mzgoddard:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/reference-script-template/build/index.html)

Close #450 and #485

To improve the test page setup experience (#369), create a copy of the reference page during the test generation for each setup script. This copy will embed the setup script and a button to activate that script once. After running the test the window can be refreshed with `Ctrl + R`, `Cmd + R`, `F5`, etc. As the script and button are now part of the reference page the test instruction page does not to try and modify it at the right time.

- use TestWindow in review page to open the reference page when Open Test Page is clicked
- update TestWIndow to not run a setup script and change its behavior to properly re-enable Open Test Page button even after reloading the reference page.
- create a template from a test plan's reference page and render a script and button into it
  - use a event listener on document to listen for a click on the button to bubble up to run the setup script
- restructure createExampleTests
  - make createExampleTests into an async function
  - add readCSV that returns a promise resolving to an array of csv rows.
  - replace csv stream callback waterfall with an await on all of the readCSV promises
  - add FileRecord filesystem helpers to recursively read test plan input and structure the output and coordinate changes to the disk
  - make changes to the destination build directory in one group of actions if the script is not validating

## TODO

- [x] Use the rendered reference page in review pages


howard-e (approved):

lib/data/parse-support.js
```diff
@@ -0,0 +1,28 @@
+/// <reference path="../../types/aria-at-csv.js" />
```
alflennik, line 1: I'm seeing several empty files, lib/data/csv.js, lib/data/file.js and lib/data/parsed.js - do you know what that's about?

alflennik (approved):
It's a delight to see the code get cleaner with each iteration!

The ARIA-AT App should start showing the setup button on the new imported test plan versions.

lib/data/parse-support.js
mzgoddard: Ah. I'll remove those. They were a step in a process I want to continue moving functions out of `scripts/create-example-tests.js`

PR #473, opened 2021-09-01
Move test plan titles for app from support.json to references.csv

howard-e:
[Preview Tests](https://raw.githack.com/w3c/aria-at/issue-471/build/index.html)

Addresses https://github.com/w3c/aria-at/issues/471.

Removed `examples` from `<root>/tests/support.json`. Moved existing titles to `<root>/tests/<directory>/data/references.csv`

Note: https://github.com/w3c/aria-at-app/pull/320 approval is dependent on this PR.

PR #521, opened 2021-10-05
Add title reference field to each test plan

mzgoddard:
#473 move titles for each test plan from support.json into their references.csv.

This PR adds a tittle to test plans that did not have one.

### DIFF
--- a/tests/menu-button-actions/data/references.csv
+++ b/tests/menu-button-actions/data/references.csv
@@ -1,6 +1,7 @@
 refId,value
 author,Isabel Del Castillo
 authorEmail,isa.delcastillo5@gmail.com
+title,Action Menu Button Example
 reference,reference/2021-1-14_12348/menu-button-actions.html
 designPattern,https://w3c.github.io/aria-practices/#menubutton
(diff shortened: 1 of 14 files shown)


tests/menu-button-actions/data/references.csv
```diff
@@ -1,6 +1,7 @@
 refId,value
 author,Isabel Del Castillo
 authorEmail,isa.delcastillo5@gmail.com
+title,Action Menu Button Example
```
jscholes, line 4: This should be "`Action Menu Button Example Using element.focus()`"

tests/toggle-button/data/references.csv
```diff
@@ -1,8 +1,9 @@
 refId,value
 author,Isabel Del Castillo
 authorEmail,isa.delcastillo5@gmail.com
+title,Toggle Button with aria-pressed
```
jscholes, line 4: I don't think `aria-pressed` is necessary here.  Just "Toggle Button".

jscholes (changes requested):
@mzgoddard Thanks for doing this!  I left some individual line comments.

PR #575, opened 2021-11-16
fix KeysInput.fromCollectedTest

mzgoddard:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/fix-keys-input-from-collected-test/build/index.html)

Use keystrokes stored in collected test keypresses member.


alflennik (approved):
I just verified this fix on the app, and double checked that the "Preview Tests" still works. Looks great. Thanks so much Z!

Issue #589, opened 2022-01-03
Error message is unclear when key constant(s) is/are missing


jscholes:
When building tests where `commands.csv` contains identifiers that are not defined in `tests/resources/keys.mjs`, the appropriate error handling seems to be missing, and hence the output of the build job is unclear and doesn't help a user solve the problem.  The message is:

> Cannot destructure property 'keystroke' of 'undefined' as it is undefined.

The run then continues, but fails at the review step because `commands.json` isn't successfully generated.  The system should inform a test writer of:

1. the testing tasks with invalid commands; and
2. which commands were considered to be invalid.

It should then exit, rather than attempting to continue with a process that will never complete.  As things stand, fixing these errors in a test plan is extremely tedious.

CC @s3ththompson

jscholes:
This is also related to https://github.com/w3c/aria-at/issues/537#issuecomment-954871244 and https://github.com/w3c/aria-at/pull/536; the build process doesn't exit when encountering this problem.

PR #608, opened 2022-01-17
Provide error message when required key constant(s) in `keys.mjs` is missing

howard-e:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/fix-issue-589/build/index.html)

Address #589.

`npm run build` will now run as expected but will report the issue with a suitable error message.

Eg. `[Command]: The key reference "ENTERR" found in "tests/alert/data/commands.csv" for "test id 1: trigger alert" is invalid. Command may not be defined in "tests/resources/keys.mjs".`

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -40,7 +40,7 @@
             <td><a href="./tests/alert/index.html">Index</a></td>
             <td><a href="./review/alert.html">Review</a></td>
             <td>3</td>
-            <td><a href="https://github.com/w3c/aria-at/commit/db05bc9" target="_blank">db05bc9 Added more details to error message
+            <td><a href="https://github.com/w3c/aria-at/commit/06f0e02" target="_blank">06f0e02 use existing addCommandError function; default keypress return
 </a></td>
           </tr>
           <tr>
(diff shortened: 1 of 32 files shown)


mzgoddard (changes requested):
The call to `addCommandError` is reporting this error already. What do you think about moving this copy into `addCommandError`? Or we can change `addCommandError` in other ways.

howard-e:
> The call to `addCommandError` is reporting this error already. What do you think about moving this copy into `addCommandError`? 

@mzgoddard I went ahead and used the existing `addCommandError` and moved the copy there as suggested. I changed the copy as such: `[Command]: The key reference "ENTERR" found in "tests/alert/data/commands.csv" for "test id 1: trigger alert" is invalid. Command may not be defined in "tests/resources/keys.mjs".`

To address the issue, the `undefined` return for the keypress object (`{ id: ..., keystroke: ... }`) that happens when the key's definition is missing has been defaulted. This means the script will continue running as expected and report the error afterwards as expected.

alflennik (approved):
Huge improvement in the readability of the errors, nice. It's true the affected test will not run on the review page, but that seems fine given the fact that a command is broken.

jesdaigle (approved):

Issue #603, opened 2022-01-05
Missing error handling when setup script file does not exist


jscholes:
If a file representing a named setup script for a test in `tests.csv` doesn't exist, there is no explicit error handling, leaving only a useless JavaScript error and stack trace that doesn't inform test writers of the specific problem or how to fix it:

```
(node:4376) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'path' of undefined
    at collectTestData (L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:1160:92)
    at L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:663:7
    at Array.map (<anonymous>)
    at L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:662:27
    at Array.flatMap (<anonymous>)
    at createExampleTests (L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:661:41)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async L:\Dropbox\Dev\aria-at\scripts\create-all-tests.js:65:55
    at async Promise.all (index 9)
    at async main (L:\Dropbox\Dev\aria-at\scripts\create-all-tests.js:63:3)
```

This doesn't indicate the test plan being processed, the specific test, the name or description of the relevant setup script, the path to the file that was constructed and not found, or even the fact that a file couldn't be found.

This is also related to https://github.com/w3c/aria-at/issues/537#issuecomment-954871244 and https://github.com/w3c/aria-at/pull/536; the build process doesn't exit when encountering this problem.

CC @s3ththompson

PR #610, opened 2022-01-17
Provide error message when setup script for test plan does not exist

howard-e:
Address #603.

When running `npm run build`, an error will now be shown when a setup script file is not properly defined in `tests/<test>/tests.csv`.

Eg. `ERROR: Setup script setFocusOnButtonn found in "tests/alert/tests.csv" for "test id 2: trigger alert" not defined in "tests/alert/reference".`


scripts/create-example-tests.js
```diff
@@ -1144,9 +1147,24 @@ function validateTest(testParsed, data, { addTestError = () => {} } = {}) {
...
+) {
+  const referencePageLocation = example.where({
+    name: test.setupScript ? test.setupScript.name : '',
+  });
+
+  if (!referencePageLocation) {
+    console.error(
+      `ERROR: Setup script ${test.setupScript.name} found in "${testPlanDirectory}/tests.csv" for "test id ${test.testId}: ${test.task}" not defined in "${testPlanDirectory}/reference".`
+    );
+    process.exit(1);
```
mzgoddard, line 1165: I don't think this error is critical enough to stop the process. Any error recorded by `addTestError`, `addCommandError`, etc, will be reported together at the end of processing this test plan, an `isSuccessfulRun: false` will be reported to `create-all-tests`, which in response will include its failure in the summary and `exit(1)`.

scripts/create-example-tests.js
```diff
@@ -1144,9 +1147,24 @@ function validateTest(testParsed, data, { addTestError = () => {} } = {}) {
  * @param {Queryable<AriaATParsed.Key>} data.key
  * @param {Queryable<{name: string, path: string}>} data.example
  * @param {Queryable<{at: string, mode: string, render: function({key: *}): string}>} data.modeInstructionTemplate
+ * @param {string} testPlanDirectory
  * @returns {AriaATFile.CollectedTest}
  */
-function collectTestData({ test, command, key, example, modeInstructionTemplate }) {
+function collectTestData(
+  { test, command, key, example, modeInstructionTemplate },
+  testPlanDirectory
```
mzgoddard, line 1155: I think we should keep file path information outside of these functions (`collectTestData`, `validateTest`, `parseCommandCSVRow`, etc). We can add path information in the main function as part of defining `addTestError` and `addCommandError` for example.

scripts/create-example-tests.js
```diff
@@ -1144,9 +1147,24 @@ function validateTest(testParsed, data, { addTestError = () => {} } = {}) {
...
+) {
+  const referencePageLocation = example.where({
+    name: test.setupScript ? test.setupScript.name : '',
+  });
+
+  if (!referencePageLocation) {
+    console.error(
+      `ERROR: Setup script ${test.setupScript.name} found in "${testPlanDirectory}/tests.csv" for "test id ${test.testId}: ${test.task}" not defined in "${testPlanDirectory}/reference".`
```
mzgoddard, line 1163: There  is an existing error for this in `validateTest`.

```js
addTestError(`Setup script does not exist: ${testParsed.setupScript.name}`);
```

mzgoddard (changes requested):
Hmm. I think there is an existing error produced for this.

```js
if (testParsed.setupScript && !data.script.where({ name: testParsed.setupScript.name })) {
  addTestError(`Setup script does not exist: ${testParsed.setupScript.name}`);
}
```

The scripts are the files in `data/js` of a test plan. For each script an example page is created. As such I think the existing error covers this.

We could use something like your copy to improve that error. And we can add the file path information to `addTestError` instead of adding it to these data processing functions. Do you think that'd work?

scripts/create-example-tests.js
howard-e: Reused the existing function and updated the copy as `[Test 1]: Setup script does not exist: "setFocusOnButtonn" for "test id 1: trigger alert" not defined.`

alflennik (approved):
Great to see readable errors for what should be a pretty common situation.

Issue #593, opened 2022-01-04
Missing error handling for invalid reference rows


jscholes:
When building tests, error handling is not implemented to inform test writers when an invalid row is encountered in `references.csv`, and assist them in fixing the problem.  For example, this row has three columns instead of two (the URL is accidentally repeated):

```
none,https://w3c.github.io/aria/#none,https://w3c.github.io/aria/#none
```

Upon attempting to parse this row, the build system outputs:

> TypeError: Cannot read property 'trim' of undefined

This doesn't indicate what the problem was, which reference row it related to, or even that there is a problem with the references file at all.

This is also related to https://github.com/w3c/aria-at/issues/537#issuecomment-954871244 and https://github.com/w3c/aria-at/pull/536; the build process doesn't exit when encountering this problem.

CC @s3ththompson

SamuelHShaw:
@s3ththompson Hi, just checking in from PAC to see if there are any updates on this, and possibly an ETA on when it may be able to be addressed?

PR #613, opened 2022-01-24
Provide error message for invalid rows in references.csv

howard-e:
Address #593.

An error message is also shown for a missing value as such: `ERROR: value not found for refId "none" in "tests/vertical-temperature-slider/data/references.csv"`


scripts/create-example-tests.js
```diff
@@ -593,14 +593,24 @@ ${rows}
   ]);
 
   for (const row of refRows) {
-    refs[row.refId] = row.value.trim();
+    // assume refId and value column exists
+    const key = row.refId || row[0];
```
howard-e, line 597: Feels a bit hacky but not seeing where we could put the logic other than `readCSV`.
Any thoughts on the original issue here @mzgoddard?

scripts/create-example-tests.js
```diff
@@ -593,14 +593,24 @@ ${rows}
...
+    const value = row.value || row[1];
+
+    if (!value) {
+      log.warning(
```
mzgoddard, line 601: Use `log.error`.

scripts/create-example-tests.js
```diff
@@ -593,14 +593,24 @@ ${rows}
   ]);
 
   for (const row of refRows) {
-    refs[row.refId] = row.value.trim();
+    // assume refId and value column exists
+    const key = row.refId || row[0];
+    const value = row.value || row[1];
+
+    if (!value) {
```
mzgoddard, line 600: Would `typeof value !== 'string'` be a better test?

scripts/create-example-tests.js
mzgoddard: I think `typeof row.refId !== 'string'` should be treated as an error like with `row.value`.

mzgoddard: Hmm. Maybe instead of testing here it should be a separate function in the file scope like `validateCSVKeys`, that you add to the references csv promise chain? Maybe have it just perform tests and not modify refRows, then have the original for loop here?

and we can `.catch` on the `Promise.all(/* ... */)` call and `log.error` any error produced then.

The for loop here is legacy code for the older data file format. So I've been (probably overly) cautious in making changes to any such code.

```js
function validateCSVKeys({requiredKeys}, rows) {
  // test rows for keys ...
  // throw if a row is missing requiredKeys
  return rows;
}
```

mzgoddard (changes requested):
I think we should error if `refId` isn't a string as well as `value`.

alflennik (approved):
Looks great. The errors are way more helpful now!

jesdaigle (approved):

Issue #599, opened 2022-01-04
Non-unique link text on review page


jscholes:
On the [Test Patterns for Review](https://w3c.github.io/aria-at/build/) page, each link in the "Index Page" and "Review Page" columns relates to a specific test plan.  However, each one has the same text ("Index" and "Review", respectively).  When combined with https://github.com/w3c/aria-at/issues/598, it is impossible for a screen reader user to quickly located a link relating to a pattern of interest without always navigating down the left-hand "Pattern" column first.

#### Suggested Solution

Use a combination of `aria-label` and `aria-labelledby` to give each link accessible names like: "Index: `<plan>`" and "Review: `<plan>`".

CC @s3ththompson

SamuelHShaw:
@s3ththompson Hi, just checking in from PAC to see if there are any updates on this, and possibly an ETA on when it may be able to be addressed?

Issue #598, opened 2022-01-04
Missing row headers on test review page


jscholes:
On the [Test Patterns for Review](https://w3c.github.io/aria-at/build/) page, the column headers are correctly marked up using `<th>` elements.  However, each pattern name (in the left-hand column) is only marked up as `<td>`.  As a screen reader user moves down through a column other than the first, they aren't informed about which pattern they are navigating over, and must always move back to the first column to make sure they're on the intended one.  This is particularly frustrating if, for example, moving down the "Review Page" column, as each link has the same text and it's impossible to quickly locate the link for the pattern of interest.

CC @s3ththompson

PR #614, opened 2022-01-24
Provide unique table cell text on review page

howard-e:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/fix-issue-599/build/index.html)

Address #599.

Also related to #598 and https://github.com/w3c/aria-at/pull/611.


mzgoddard (approved):
Looks good to me.

alflennik (commented):
It seems to me that adding the `th` might be enough to solve the problem fixed with the aria-labels, so perhaps the aria-labels aren't needed. @jscholes feel free to weigh in?

jscholes:
@alflennik There are situations where table headers won't be conveyed alongside the link.  For instance, when using the links list in a screen reader (e.g. NVDA+F7, JAWS+F7).  So the link text being unambiguous is also a good idea.  However, in the case of the commit links, the text is overly verbose, and the "Commit of Last Change:" prefix should be removed.

jscholes:
@mcking65 I mostly recommend that the unique context be provided at the end of the existing control, and be separated by something which triggers a pause in screen reader speech.  So rather than "Actions where X", I would want "Actions: X", with a colon and a space.  This, in my view:

1. doesn't add too much cognitive overhead, because the information is easy to ignore given its position and separation; and
2. facilitates the use case of distinguishing between the controls with certain screen reader features, like the JAWS/NVDA+F7 elements list.

It also provides the maximum amount of flexibility, because table headers can be turned off, without losing context.  I don't think it's perfect, for a number of reasons, including the fact that row and column headers usually can't be turned off independently of each other.  But, it's what we have.

@howard-e Commit text is much better, thanks!

Issue #597, opened 2022-01-04
Commit hash missing from review page


jscholes:
On the [Test Patterns for Review](https://w3c.github.io/aria-at/build/) page, the "Commit of Last Change" column only contains the word "commit" for the most recent commit on the most recently modified (or merged?) test plan row.  For example, at the time of writing, the most recently merged pattern is "seek-slider", and all test plans other than that one show a commit hash as expected.  Note that this info will become out of date as additional plans are modified and merged into master; look for the row with the word "commit" instead of a hash.

This also impacts pull requests.  For instance, here's the review page for PR https://github.com/w3c/aria-at/pull/545:

https://raw.githack.com/w3c/aria-at/tests/banner/build/index.html

As you'd expect, the most recently modified test plan on that branch is "banner", and the "banner" row only has the word "commit" in the last column instead of a hash.  I'll leave that branch in place, rather than deleting it, to serve as an example for this issue.

CC @s3ththompson

SamuelHShaw:
@s3ththompson  Hi, just checking in from PAC to see if there are any updates on this, and possibly an ETA on when it may be able to be addressed?

s3ththompson:
Hi @SamuelHShaw, thanks for filing this issue! The Bocoup office was closed last week for an all-hands. We're working through these issues as we can get to them since we're currently in the middle of the Q1 2022 planning process. My understanding is that we'll have a PAC / Bocoup partner meeting soon to discuss how we can best work together and provide timely ETAs for support issues going forward.

PR #617, opened 2022-02-07
Address missing review page information

howard-e:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/fix-issue-597/build/index.html)

Address #597.

The `generate-and-commit-files` step was incorrectly referencing the master branch's head rather than the submitted PR's head.

Additionally, the commit history was defaulted to a depth of 1 so the "Commit of Last Change" was duplicated across all test plans rows which made that value wrong for most.

[See example preview link from test PR where `alert-test` is a newly created test plan](https://raw.githack.com/howard-e/aria-at/test-build-output-6/build/index.html) from [my fork.](https://github.com/howard-e/aria-at/pull/16)

### DIFF
--- a/.github/workflows/generate-and-commit-files.yml
+++ b/.github/workflows/generate-and-commit-files.yml
@@ -12,7 +12,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          ref: ${{ github.head_ref }}
+          ref: refs/pull/${{ github.event.pull_request.number }}/head
+          fetch-depth: 0
 
       - name: Use Node.js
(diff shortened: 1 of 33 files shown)


alflennik (approved):
When reviewing this PR we had some trouble due to the fact that GitHack was down. Hopefully this issue isn't too persistent or we might need to look into using another proxy, perhaps the one ARIA-AT App provides. Fortunately we were able to work around it and confirm that the fix is working. Thanks for setting up a fake test plan PR, making it easy to confirm the fix.

jesdaigle (approved):

Issue #590, opened 2022-01-03
Tests fail to build when encountering UTF-8 byte order mark


jscholes:
I'm attempting to build the tests in https://github.com/w3c/aria-at/pull/426, but the build job run is consistently failing with an unhelpful error (e.g. as seen in https://github.com/w3c/aria-at/runs/4695547690?check_suite_focus=true):

> TypeError: Cannot read property 'target' of undefined

I have no idea what this is refering to, and there is no additional context other than the stack trace:

```
(node:16656) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'target' of undefined
    at collectTestData (L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:1159:19)
    at L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:663:7
    at Array.map (<anonymous>)
    at L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:662:27
    at Array.flatMap (<anonymous>)
    at createExampleTests (L:\Dropbox\Dev\aria-at\scripts\create-example-tests.js:661:41)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async L:\Dropbox\Dev\aria-at\scripts\create-all-tests.js:65:55
    at async Promise.all (index 0)
    at async main (L:\Dropbox\Dev\aria-at\scripts\create-all-tests.js:63:3)
```

I added some debug logging in `create-example-tests.js`, and the output seems to suggest that the numeric test ID for the first task in the accordion plan isn't being detected for some reason:

```
{
  testId: NaN,
  task: 'navigate forwards to expanded accordion header',
  title: 'Navigate forwards to an expanded accordion header in reading mode',
...
```

I'm not sure why `testId` would be `NaN` here, but that seems to be the crux of the problem.  Without a test ID, the corresponding command row cannot be located.  However, I'm at a loss as to why `tests.csv` and/or `commands.csv` (the scripts don't tell me anything to help me work out which one) isn't/aren't being parsed correctly.  The column headers are set correctly (not even sure if that matters to the parsing or not), and I can't spot any issues with non-standard spacing, line breaks, etc.

Without anything more to go on, I'm effectively blocked from working on this test plan further right now.  CC @s3ththompson / @sinabahram

jscholes:
[...]

PR #618, opened 2022-02-08
Provide warning if UTF-8 byte order mark found when processing CSVs

howard-e:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/fix-issue-590/build/index.html)

Address #590.

Based on https://github.com/w3c/aria-at/issues/590#issuecomment-1021413237

> @s3ththompson 
> .. I'm not sure yet whether it's appropriate to silently ignore the BOM mark, but it does seem like we should be able to at least warn when it's encountered.

I've gone ahead and logged a warning message back to the terminal of its presence, as we're currently unsure of any other side effects it may cause (although I haven't come across any others while testing) so removing it would be ideal.

It is also removed from the CSV being processed, to allow for the script to continue running as expected.

### DIFF
--- a/build/index.html
+++ b/build/index.html
@@ -21,11 +21,6 @@
         padding: .5em;
       }
 
-      tr:not(:first-of-type) th {
-        padding: .5em;
-        text-align: left;
-      }
-
     </style>
   </head>
   <body>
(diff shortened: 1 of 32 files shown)


alflennik (approved):

alflennik (approved):
We went through this one together and I got to see it fail on master and then succeed in the branch. I like that we're not silently swallowing the error, allowing authors to come in and fix those errant characters.

jesdaigle (approved):

Issue #616, opened 2022-01-25
Invalid husky script


howard-e:
When attempting to run `git commit`, I'm presented with a message as such:

```sh
❯ git commit -m "git commit message"
npm ERR! Missing script: "test"
npm ERR! 
npm ERR! To see a list of scripts, run:
npm ERR!   npm run

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/user/.npm/_logs/2022-01-25T15_32_42_915Z-debug-0.log
husky - pre-commit hook exited with code 1 (error)
```

The `.husky/pre-commit` script includes a `npm test` which looks for that script inside `package.json`. The `test` script doesn't exist and I can only commit with the `--no-verify` flag to skip, which consequently skips the `npx lint-staged` step.

PR #620, opened 2022-02-09
Remove invalid test command from husky pre-commit script

howard-e:
Address #616.


nschonni:
@jesdaigle would be nice to get this one landed too, since it causes errors during local dev. I guess it might need a rebase though

howard-e:
> would be nice to get this one landed too, since it causes errors during local dev. I guess it might need a rebase though

As it doesn't directly impact the test writing functionality and only beneficial from what I see, I'll go ahead and rebase and merge with the single approval. Thanks for the ping there @nschonni.

PR #606, opened 2022-01-17
Create build folder if it doesn't exist when running `create-example-tests.js`

howard-e:
[Preview Tests](https://raw.githack.com/w3c/aria-at/bocoup/fix-issue-588/build/index.html)

Address #588.

Issue #588, opened 2022-01-03
`npm run build` doesn't create build directory, fails if it doesn't exist


jscholes:
I recently needed to run `npm run cleanup`, to completely remove the build directory in order to avoid merge conflicts.  This worked as expected, but when subsequently running `npm run build`, this was the result:

```
> aria-at@1.0.0 build L:\Dropbox\Dev\aria-at
> npm run create-all-tests && npm run review-tests


> aria-at@1.0.0 create-all-tests L:\Dropbox\Dev\aria-at
> cross-var node scripts/create-all-tests.js --testplan=$npm_config_testplan

The destination 'build' directory does not exist.
npm ERR! code ELIFECYCLE
```

If a command is provided to remove the build directory before rebuilding it, it seems suboptimal that rebuilding it then doesn't work without the additional step of manually recreating the directory.  Particularly because Git doesn't allow you to commit an empty directory addition by default, so it's impossible to do something like:

```
npm run cleanup
git add --all .
git commit -m "Remove build directory"
git push
```

... and have the GitHub Actions workflow actually work.

CC @s3ththompson

Issue #384, opened 2021-02-02
Potential GitHub automation workflow/script improvements


jscholes:
The GitHub Action, triggered by pushing to a PR branch, currently executes the [create-all-tests](https://github.com/w3c/aria-at/blob/master/scripts/create-all-tests.js) and [review-tests](https://github.com/w3c/aria-at/blob/master/scripts/test-reviewer.mjs) scripts.  These cause all assets to be regenerated for all patterns present in the `tests` directory, even though a test developer will typically only be working on one pattern at a time.

As the number of patterns in the `tests` directory increases, so will the running time of the scripts and workflow.  The current approach also increases the likelihood of merge conflicts (e.g. like those seen in PR #330), because files are updated which have no relation to the test pattern under development.

#### Additional Observations

1. The GitHub workflow runs indiscriminately, regardless of context.  For example:
    * If a developer manually runs the [create-tests script](https://github.com/w3c/aria-at/blob/master/scripts/create-tests.js) locally, to only regenerate assets for a specific pattern, the create-all-tests script will run anyway on push.
    * If a pull request is submitted which has nothing to do with test writing (e.g. an editorial update to the README or other unrelated file(s)), all tests will be regenerated and a preview link added to the PR.
2. All pattern assets are regenerated in full, irrespective of which changes have been made since the last generation.  E.g. files for all 76 "Editable Combobox With Both List and Inline Autocomplete " tests are recreated, even if a small change is made to only one of the rows in `tests.csv`.

#### Suggested Enhancements

In priority order:

1. Automatically detect, or allow a test developer to manually specify, the applicable patterns on a particular branch, and only regenerate assets for those patterns.
    * This ties in with issue #371 (enhancements for PR preview links).
    * If a PR requires all tests to be regenerated for some reason, developers should have a way of specifying such.
    * If a PR contributor doesn't specify some or all patterns for regeneration, the GitHub workflow shouldn't run.
2. Detect when a developer has already generated test assets locally, to avoid unnecessary execution of the GitHub workflow and the `git push` followed by `git pull` dance.
3. Implement a differential strategy to avoid regenerating all test assets when only small, targeted changes have been made.

s3ththompson:
Here are the proposed steps we discussed in the last CG meeting:

1. Put verbose logging behind flag
2. Summarize errors with better line & file explanations
3. Add ability to validate only
4. Add ability to build/validate single test plan given as argument
5. Turn on Netlify continuous deployment and PR deploy previews
[...]

PR #592, opened 2022-01-04
chore: migrate nodegit to simple-git

nschonni:
[Preview Tests](https://deploy-preview-592--aria-at.netlify.app)

Found that the NPM install was stalling from this dependency so swapped with simple-git https://www.npmjs.com/package/simple-git

Husky/lint-staged changes were because the hooks are currently broken, as there is no `npm test` script

PR #621, opened 2022-02-15
Remove build folder and use Netlify deploys for previews

howard-e:
There has been some discussion around using Netlify to present PR previews (and by extension, serve as the published site because the repo's [current site link](https://w3c.github.io/aria-at/build/index.html) depends on the build folder, which we are now proposing to remove).

Previous discussion can be seen at:
* https://github.com/w3c/aria-at/issues/384
* https://github.com/w3c/aria-at/pull/441#issue-910668769

This PR introduces Netlify's use by:
1. untracking the [build](https://github.com/w3c/aria-at/tree/master/build) folder,
1. including `netlify.toml` with the directives required to display the site's content through Netlify,
1. updating the site preview link used in PRs to point to the generated netlify deploy link, and
1. ensuring the site's build folder can be created consistently without depending on a pre-existing build folder as reported in https://github.com/w3c/aria-at/issues/588 and fixed (pending merge) in https://github.com/w3c/aria-at/pull/606.

Assumptions made here:
* The site's name will be `aria-at`, and so the site's url will be https://aria-at.netlify.app.
* Netlify 'Builds' are active.
* Netlify 'Deploy Previews' feature is enabled.
* Netlify 'Drawer' feature is disabled.

Changes as a result of this:
* There should be no change to how contributions are currently done.
* `Generate and Commit Files` workflow will need to be disabled.
* This should hopefully make reviews a little easier to see what changed without having to filter through generated files.
* This should hopefully be more reliable than raw.githack.com's URLs (we have recently been receiving `429: Too Many Requests` errors when viewing PR Previews)

___

**NOTE**
* Since there is 5000+ file changes in this commit due to the build folder removal, you can view https://github.com/w3c/aria-at/commit/fa246913593d6ac8b65cfef0aec5ba87c44284cd to see the files that changed.
* You can view a deployed [site example here](https://deploy-preview-8--aria-at-howarde.netlify.app) along with the PR which triggered this build - https://github.com/howard-e/aria-at/pull/8.


howard-e:
Note that I manually cancelled the `Generate and Commit Files` workflow as it would be unnecessary here and the `Update Pull Request with Preview Link` isn't able to run from a forked repository, to explain the build failures.

jscholes:
@howard-e Workflow question: currently, the most common way we check that a build can be successfully performed is to just push up to GitHub, and let the action take care of it.  Will that still be possible with the proposal outlined here?  I.e. will the status of the build still be reported on the relevant PR?

jesdaigle (approved):
This is amazing. Thank you @howard-e !

jscholes:
@howard-e I'm open to whatever.  If we should just validate the tests locally instead, that's fine as a workflow change.  Might actually be easier.

jesdaigle:
@mcking65 @jscholes Merging this ticket is dependent on having a Netlify account that has the right credentials to this repo. It's currently using Howard's personal account.

nschonni:
https://github.com/w3c/aria-at/pull/592 might resolve the need to pin to Node 12, since it's the native build that appears to be failing on a few of the other PRs that recently tried to run Netlify

howard-e:
> #592 might resolve the need to pin to Node 12, since it's the native build that appears to be failing on a few of the other PRs that recently tried to run Netlify

@nschonni yep, it might. This looks good to go now so let's land this one and test with that after.

Wanted to also maintain parity with the node version that's been pinned in the workflows but that may have been the reason why it was pinned as well.

PR #632, opened 2022-02-17
chore: build but don't push for Dependabot

nschonni:
Attempt 3 ...
Run the build like it currently since the skip isn't working, but don't push the result. This will at least test if the generate script works with any package changes.
Move the skiping to the lint job to avoid running it on push and PR, but try doublequotes for the globs


nschonni:
@howard-e maybe this time 😄

howard-e (approved):

PR #643, opened 2022-03-15
ci: use pull_request_target for PR preview

nschonni:
needed since it requires the GH token to update the issue/PR through the API


howard-e (approved):
LGTM

PR #644, opened 2022-03-16
Ensure scripts rootDirectory is consistent

howard-e:
This is an issue which may arise when contents of aria-at has to be processed from another environment such as https://github.com/w3c/aria-at-app where an additional 'scripts' may be a part of the path.

### DIFF
--- a/scripts/create-all-tests.js
+++ b/scripts/create-all-tests.js
@@ -38,7 +38,7 @@ async function main() {
   const VALIDATE_CHECK = !!args.validate;
 
   const scriptsDirectory = path.dirname(__filename);
-  const rootDirectory = scriptsDirectory.split('scripts')[0];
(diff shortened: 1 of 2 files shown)


scripts/create-all-tests.js
```diff
@@ -38,7 +38,7 @@ async function main() {
   const VALIDATE_CHECK = !!args.validate;
 
   const scriptsDirectory = path.dirname(__filename);
-  const rootDirectory = scriptsDirectory.split('scripts')[0];
+  const rootDirectory = path.join(scriptsDirectory, '../');
```
nschonni, line 41: Don't think you need the trailing `/`, just the `..`. probably better for cross environment too

howard-e: Thanks for that note!

howard-e: Done

PR #646, opened 2022-03-16
Generate html source script to support aria-at-app

howard-e:
[Preview Tests](https://deploy-preview-646--aria-at.netlify.app)

https://github.com/w3c/aria-at-app's TestRenderer depends on test example html artifacts which contains scripts. This previously existed in the `build` folder, now removed by #621. This PR adds those required artifacts without the need to commit the build folder.

`npm run build` already creates these files in the (now, local) build folder. They're copied over to the source folder location of the example, ie. `tests/<testplan>/reference/<path>/<testplan>.<script>.html`.

These files are updated by:
* Running `npm run build` before commit and push, OR
* `Generate and commit files` workflow runs `npm run build`, and has had the language updated to reflect the change. (**This will not run if there are no updated files and it will not cause unintended merge conflicts through clashing generated files as previously experienced; line endings for those files are ignored to help ensure this)**
* `npm run build` will also remove the html script artifacts which are no longer being generated (most likely due to script being removed from `tests/<testplan>/data/js/<script>.js`

cc @s3ththompson @richnoah


alflennik (approved):
I'll definitely be curious if these can be removed in the future, but based on our discussion I think this is a super clean solution, and still allows us to delete hundreds of thousands of lines of generated files from the repo. And these files are way less volatile and will rarely, if ever, change. Thanks for getting us through that huge forward leap Howard!

PR #649, opened 2022-03-21
chore: update package lock v2 and Node 16

nschonni:
[Preview Tests](https://deploy-preview-649--aria-at.netlify.app)

Bump the packlock to v2 and Node 16 running now that the node-git block is resolved. Ran `npm audit fix` in separate commits after the original automated conversion

### DIFF
--- a/.github/workflows/generate-and-commit-files.yml
+++ b/.github/workflows/generate-and-commit-files.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: '12.x'
+          node-version: '16.x'
           cache: 'npm'
 
       - name: Install dependencies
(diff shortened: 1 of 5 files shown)


howard-e:
Thanks @nschonni! LGTM!

Just curious if there could be any adverse effects to this that we may need to be aware of @mzgoddard? I vaguely remember node version related issues popping up before around the build process. But I can't seem to track down the conversations related to those items to verify.

howard-e (approved):
Looks good @nschonni

> Just curious if there could be any adverse effects to this that we may need to be aware of? I vaguely remember node version related issues popping up around the build process. But I can't seem to track down the conversations related to those items to verify.

Nevermind @mzgoddard! Found the [issue](https://github.com/w3c/aria-at/pull/499) I was referring to. All looks good.

nschonni:
Oh, I think lint-staged can now bump with this too, but that can always be done separately

howard-e:
> Oh, I think lint-staged can now bump with this too, but that can always be done separately

Let's do that separately. Will land this.

PR #666, opened 2022-03-23
Remove `htmlparser2` dependency

howard-e:
[Preview Tests](https://deploy-preview-666--aria-at.netlify.app)

Remove `htmlparser2` dependency as it's no longer being used.

Related to https://github.com/w3c/aria-at/pull/656#issuecomment-1076509042.


mzgoddard (approved):
Looks good to me. I think this was last being used in the test generation script but that is no longer the case.

PR #671, opened 2022-03-29
chore: add eslint-plugin-n

nschonni:
[Preview Tests](https://deploy-preview-671--aria-at.netlify.app)

Splitting out from other PR to see if this change only is OK with the build

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,4 @@
 tests/**/reference/**/*
 build/tests/**/*
+# keys are manually parsed in create-example-test.js and depend on format
+tests/resources/keys.mjs
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,6 +6,7 @@
   },
   "extends": [
     "eslint:recommended",
+    "plugin:n/recommended",
     "plugin:import/errors",
     "plugin:import/warnings",
     "plugin:prettier/recommended"
@@ -15,12 +16,13 @@
     "SharedArrayBuffer": "readonly"
   },
   "parserOptions": {
-    "ecmaVersion": 2018,
+    "ecmaVersion": 2021,
     "sourceType": "module"
   },
   "rules": {
     "no-dupe-keys": 0,
     "no-undef": 0,
-    "no-unused-vars": 0
+    "no-unused-vars": 0,
+    "n/no-process-exit": 0
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "cross-var": "^1.1.0",
         "csv-parser": "^3.0.0",
+        "eslint-plugin-n": "^15.1.0",
         "fs-extra": "^10.0.1",
         "inquirer": "^8.2.2",
         "json-beautify": "^1.1.1",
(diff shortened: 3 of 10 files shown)


nschonni:
@howard-e I split off the new plugin from the other PR to try an narrow down the failure, but I'm not sure why this one that really just reformats the `.mjs` files is hitting the `.trim` error

howard-e:
> @howard-e I split off the new plugin from the other PR to try an narrow down the failure, but I'm not sure why this one that really just reformats the `.mjs` files is hitting the `.trim` error

@nschonni it seems https://github.com/nschonni/aria-at/blob/3b9576a823b09230d1d2db86e835adc304333686/scripts/create-example-tests.js#L180 depends on `keys.mjs` having double quotes around the values. I think that should be changed to:
```javascript
let parts2 = line.split('\'');
```

tests/resources/keys.mjs
```diff
@@ -1,131 +1,134 @@
...
+export const CTRL_OPT_UP = 'Control+Option+Up';
+export const CTRL_OPT_DOWN = 'Control+Option+Down';
+export const CTRL_OPT_RIGHT_AND_CTRL_OPT_LEFT = 'Control+Option+Right / Ctrl+Option+Left';
+export const CTRL_OPT_A = 'Control+Option+A';
+export const CTRL_OPT_CMD_J = 'Control+Option+Command+J';
+export const CTRL_OPT_CMD_L = 'Control+Option+Command+L';
+export const CTRL_OPT_CMD_P = 'Control+Option+Command+P';
+export const SHIFT_CTRL_OPT_CMD_J = 'Shift+Control+Option+Command+J';
+export const SHIFT_CTRL_OPT_CMD_L = 'Shift+Control+Option+Command+L';
+export const CTRL_OPT_CMD_J_AND_SHIFT_CTRL_OPT_CMD_J =
```
howard-e, line 30: Seems this will also need be on a single line

tests/resources/keys.mjs
```diff
@@ -1,131 +1,134 @@
...
+  'Control+Option+Command+J / Shift+Control+Option+Command+J';
+export const CTRL_OPT_CMD_C_AND_SHIFT_CTRL_OPT_CMD_C =
```
howard-e, line 32: Same as previous comment

tests/resources/keys.mjs
```diff
@@ -1,131 +1,134 @@
...
+  'Control+Option+Command+C / Shift+Control+Option+Command+C';
+export const CTRL_OPT_F3 = 'Control+Option+F3';
+export const CTRL_OPT_F4 = 'Control+Option+F4';
+export const CTRL_OPT_SPACE = 'Control+Option+Space';
+export const CTRL_OPT_SPACE_THEN_CTRL_OPT_RIGHT =
```
howard-e, line 37: Same as previous comment

nschonni:
@howard-e good catch! I'm wondering why it's manually parsing the file instead of `import * as keys from './keys.mjs';` like in the other files

howard-e:
> I'm wondering why it's manually parsing the file instead of `import * as keys from './keys.mjs';` like in the other files

Can't say I'm too sure on that but if I had to guess, because of how it's being used with the `FileRecord` and related classes to make use of file related utility methods that provides. Seems to also be consistent with how other files are being handled as well. As well as preserving how things were done before that utility to maintain how it worked before.

nschonni:
lol, took the easy way out and ignored keys.mjs from ESLinting 😆

howard-e (approved):

PR #672, opened 2022-03-30
chore: remove backup file

nschonni:
[Preview Tests](https://deploy-preview-672--aria-at.netlify.app)

Bak files  shouldn't be needed, since old files can be retrieved from source control


howard-e (approved):
Agreed!

PR #341, opened 2020-11-20
chore: Add ESLint Plugins

nschonni:

PR #673, opened 2022-03-30
Replace fs.rmSync() usage with rimraf.sync()

howard-e:
[Preview Tests](https://deploy-preview-673--aria-at.netlify.app)

There is implicit support for `node >=12` in this project. Removing fs.rmSync() which has support on [node >=v14.14.0](https://nodejs.org/api/fs.html#fsrmsyncpath-options) helps preserve this.

Based on https://github.com/w3c/aria-at/pull/341#issuecomment-1082221667.

cc @nschonni

### DIFF
--- a/scripts/create-example-tests.js
+++ b/scripts/create-example-tests.js
@@ -15,6 +15,7 @@ const {
 
 const csv = require('csv-parser');
 const beautify = require('json-beautify');
+const rimraf = require('rimraf');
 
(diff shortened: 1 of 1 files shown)


nschonni:
Since v12 is EOL at the end of April https://github.com/nodejs/Release I think it's OK. CI is already moved to v16

howard-e:
@nschonni thanks for pointing that out. Closing given that consideration

PR #702, opened 2022-04-26
Create updated tests for APG design pattern example: Editable Combobox With Both List and Inline Autocomplete

IsaDC:
Update APG example to the latest version.


IsaDC:
@jscholes  Please take a look at the scripts.

PR #707, opened 2022-04-28
Create updated tests for APG design pattern example: Main landmark

IsaDC:
[Preview Tests](https://deploy-preview-707--aria-at.netlify.app)

Update APG example to the latest version.


jscholes (changes requested):
@IsaDC For test 7, Navigate forwards out of a main landmark in reading mode, I'm making the following changes:

* add comma to the NVDA commands;
* add Shift+Period then Down Arrow to the JAWS commands; and
* add an additional press of Down Arrow within the existing JAWS commands, i.e. Down Arrow then Down Arrow.

Issue #714, opened 2022-05-16
Wording in Instruction Step 4 in task "Navigate backwards to a menu button in reading mode" (Navigation Menu Button, Test 4)


AlyssaGourley:
## Description of Behavior
In step 4 under "Test Instructions, the instruction says "With the reading cursor on the 'Navigate forwards from here' link..." Since we're navigating backwards, we should instead be focusing first on the "Navigate backwards from here" link, which is where setup sets the focus.

## Test Setup

- Test File at Exact Commit: [test-04-navigate-backwards-to-a-menu-button-reading-jaws.collected.html](https://aria-at.w3.org/aria-at/7ec8fa91c8bde5af1fe7ff78698fcf1c737af1b2/build/tests/menu-button-navigation/test-04-navigate-backwards-to-a-menu-button-reading-jaws.collected.html)
- AT: JAWS (version Latest)
- Browser: Chrome (version Latest)

AlyssaGourley:
This also occurs on the "Navigate backwards to a menu button in interaction mode" test.

jscholes:
@AlyssaGourley Good catch!  Thanks for this; all instances should now be fixed.

PR #734, opened 2022-05-25
Correct link name for Navigation Menu Button:

IsaDC:
[Preview Tests](https://deploy-preview-734--aria-at.netlify.app)

Corrected the link's name for tasks 4, 5 and 6 to reflect the intended direction.


IsaDC:
Solves #714

PR #832, opened 2022-10-06
Increase number of supported commands for tests

howard-e:
[Preview Tests](https://deploy-preview-832--aria-at.netlify.app)

This adds support for up to 26 commands in the `tests/<example>/data/commands.csv` files.


s3ththompson (approved):
Thanks for this change @howard-e!

cc @jscholes @SamuelHShaw we're ready to merge this, but I wanted to flag the following information for you before we do so:

This PR raises the previous limitation of 6 commands per test to 26 (`commandA` through `commandZ`). When reviewing, we noticed that this new behavior affects the existing breadcrumb test plan: https://github.com/w3c/aria-at/blob/master/tests/breadcrumb/data/commands.csv

Breadcrumb has a test titled "Navigate to the first breadcrumb link" in reading mode on NVDA that has a DOWN command in column `commandG`. Previously, this column was silently ignored. Now that we support additional columns, this command is visible in the [generated test html](https://deploy-preview-832--aria-at.netlify.app/tests/breadcrumb/test-01-navigate-to-the-first-breadcrumb-link-reading.html?at=nvda).

This behavior is probably desired! But I wanted to flag this change for you just to double check. Please let us know if you're okay with with merging. This should be a clean rebase if you want to add it to any existing open PRs.

jscholes (approved):

PR #904, opened 2023-03-07
Fix open file limit graceful dependency

### LINKED ISSUE #890: Parallelization exceeds system's "open file" limit
The `npm run create-all-tests` command consistently fails on my machine when it attempts to concurrently open more files than my operating system supports.

Output of `npm run create-all-tests -- --verbose`:

    npm WARN config init.license Use `--init-license` instead.
    
    > aria-at@1.0.0 create-all-tests
    > cross-var node scripts/create-all-tests.js --testplan=$npm_config_testplan
    
    node:internal/process/promises:279nfig:load:flatten Completed in 3ms
                triggerUncaughtException(err, true /* fromPromise */);
                ^
    
    [Error: EMFILE: too many open files, open '/home/mike/projects/bocoup/facebook-aria/aria-at/build/tests/checkbox-tri-state/test-03-navigate-forwards-to-a-mixed-checkbox-interaction.json'] {
      errno: -24,
      code: 'EMFILE',
      syscall: 'open',
      path: '/home/mike/projects/bocoup/facebook-aria/aria-at/build/tests/checkbox-tri-state/test-03-navigate-forwards-to-a-mixed-checkbox-interaction.json'
    }

Node version: 16.17.0 (this also occurs in 14.16.0)  
Platform: Ubuntu 18.04 (default "open file" limit: 1024)

mzgoddard:
[Preview Tests](https://deploy-preview-904--aria-at.netlify.app)

Fix #890 

Call `fs-extra` functions to work around the 'too many open files' error some systems may encounter running `create-all-tests.js`.

`test-reviewer.mjs` and `update-reference.js` use a lot of `Sync` `fs` calls, that `fs-extra` cannot work around (this is a limitation of `graceful-fs` that `fs-extra` depends on).


jugglinmike:
Can you say a bit about why you chose not to use the simpler of the three solutions I described in gh-890?

mzgoddard:
> Can you say a bit about why you chose not to use the simpler of the three solutions I described in [gh-890](https://github.com/w3c/aria-at/issues/890)?

Sure. Briefly, I don't think the first and third options should be used to solve this issue. I am curious what you think of a possible 4th option in-place of this PR's change.

I think the first option (running test plan creation in series instead of parallel) delays the issue. As of https://github.com/w3c/aria-at/pull/904/commits/59b662666f40ecba10e86a546ff09a1cc74cc26a `build` is 211 directories and 5405 files. My shell reports an unmodified file descriptor limit (with `ulimit -n`) of 4864. The current largest test plan build is `combobox-autocomplete-both-updated` with 444 files. It has 76 tests. If we use that for a (very) rough ratio of output files to tests a test plan with ~833 tests with the current build system could open enough files on its own to encounter this error on my current computer.

Documenting how to troubleshoot it, I think is kind of odd. I think this is subjective but documentation on troubleshooting or working around issues should be on matters that are rare, hard to reproduce, or the effort to fix is too high. This issue I think has low enough effort solutions and for host systems that encounter it, it should be a regular problem.

I considered a fourth option: to replace all or some of the fs calls with their synchronous versions. If we replace some of calls, like the `readFile` and `writeFile` in `file-record.js`, we would greatly limit the number of open file descriptors. It wouldn't depend on an external dependency. And I think most personal computers built in the last decade wouldn't experience a change in performance. Trying it locally my system didn't experience a difference in performance. As I understand some quick profiles on the script, most of the time is spent handling the data and producing the output content. Some computers like servers may use slow enough drive configurations to cause a noticeable slow down but I think that may be acceptable.

jugglinmike:
Hey @mzgoddard, the build script is still failing after reaching open file limits on my system, even with this patch applied.

<details>
  <summary>Sample output</summary>

    $ git show --oneline --no-patch
    10bc8929 (HEAD, upstream/fix-open-file-limit-graceful-dependency) Merge branch 'master' of github.com:w3c/aria-at into fix-open-file-limit-graceful-dependency
    $ git status
    HEAD detached at 10bc8929
    nothing to commit, working tree clean
    $ npm run build
    npm WARN config init.license Use `--init-license` instead.
    
    > aria-at@1.0.0 build
    > npm run create-all-tests && npm run review-tests
    
    npm WARN config init.license Use `--init-license` instead.ted in 2ms
    
    > aria-at@1.0.0 create-all-tests
    > cross-var node scripts/create-all-tests.js --testplan=$npm_config_testplan
    
    node:internal/process/promises:279nfig:load:flatten Completed in 3ms
                triggerUncaughtException(err, true /* fromPromise */);
                ^
    
    [Error: EMFILE: too many open files, open '/home/mike/projects/bocoup/facebook-aria/aria-at/build/tests/menu-button-actions/test-10-open-menu-reading-jaws.collected.json'] {
      errno: -24,
      code: 'EMFILE',
      syscall: 'open',
      path: '/home/mike/projects/bocoup/facebook-aria/aria-at/build/tests/menu-button-actions/test-10-open-menu-reading-jaws.collected.json'
    }

</details>

Maybe this comes from [the patch's continued use of the "fs-extra" package](https://github.com/w3c/aria-at/blob/7a8454bca6de980199868101431817cea03cce35/scripts/update-reference.js#L2).

jugglinmike (changes requested):
(see [my previous comment](https://github.com/w3c/aria-at/pull/904#issuecomment-1468903825))

mzgoddard:
@jugglinmike ah no so its not `fs-extra`. According to the second line of their readme, they use `graceful-fs`.

> It also uses [graceful-fs](https://github.com/isaacs/node-graceful-fs) to prevent EMFILE errors. It should be a drop in replacement for fs.

Instead its ... `graceful-fs`. So I'm very very surprised to find that `graceful-fs` does not patch `fs.promises`. Here's an issue https://github.com/isaacs/node-graceful-fs/issues/160 opened in 2019 for it. That issue is referenced by another https://github.com/jprichardson/node-fs-extra/issues/977 that made me wonder if `fs.promises` may have added some kind of graceful option. There is sort of one, but only for [`rm and rmdir`](https://nodejs.org/dist/latest-v18.x/docs/api/fs.html#fspromisesrmpath-options).

I'm going to push a commit to replace `graceful-fs` in this PR with more calls to `fs-extra`. It [patches `fs` methods with `universalify`](https://github.com/jprichardson/node-fs-extra/blob/0e7de32aac6689ea61b50a3b0627bfdf99444429/lib/fs/index.js#L4) letting us call `fs.readFile` and `fs.writeFile` [without callbacks to return promises](https://github.com/jprichardson/node-fs-extra#sync-vs-async-vs-asyncawait).

mzgoddard:
I updated the PR body to represent the change to use `fs-extra`.

jugglinmike:
Hey @mzgoddard, I discovered a bug on the main development branch while reviewing your work here: gh-910. I think we should address the underlying problem before moving forward with this patch (though I don't yet know whether it's appropriate to take that on here).

Would you mind taking a look?

mzgoddard:
> Hey @mzgoddard, I discovered a bug on the main development branch while reviewing your work here: [gh-910](https://github.com/w3c/aria-at/issues/910). I think we should address the underlying problem before moving forward with this patch (though I don't yet know whether it's appropriate to take that on here).
> 
> Would you mind taking a look?

I took a look and I made two PRs that independently fix the issue. Thanks for filing the issue on that.

jugglinmike (approved):
With the non-determinism reported in gh-910 resolved, I was able to vet this patch.

First, following [some extra system administration steps](https://superuser.com/questions/1200539/cannot-increase-open-file-limit-past-4096-ubuntu) (i.e. the steps we're trying to avoid requiring from new contributors), I ran the build process from the main development branch:

    $ npm run build
    $ find build -type f -print0 | xargs -0 cat | sha1sum
    858136a19e78fa2235edc64aadc1e28cc057be52  -

With that, I could verify that this change does not impact the contents of the generated files:

    $ git checkout fix-open-file-limit-graceful-dependency
    $ npm install
    $ for i in $(seq 5); do npm run build; mv build build-${i}; done
    $ for i in $(seq 5); do find build-${i} -type f -print0 | xargs -0 cat | sha1sum; done
    858136a19e78fa2235edc64aadc1e28cc057be52  -
    858136a19e78fa2235edc64aadc1e28cc057be52  -
    858136a19e78fa2235edc64aadc1e28cc057be52  -
    858136a19e78fa2235edc64aadc1e28cc057be52  -
    858136a19e78fa2235edc64aadc1e28cc057be52  -

Crucially, the revised build process completes successfully on my system *without* those extra administration steps.

Nice work, @mzgoddard!

PR #907, opened 2023-03-09
Create tests root record

mzgoddard:
[Preview Tests](https://deploy-preview-907--aria-at.netlify.app)

This change makes some changes I thought would be good to implement while working on #904.

- reduce repeated calls of `path.dirname` and `path.basename` on the same arguments
  - of note is `path.basename(directory)`
- base `existingBuild` and `newBuild` on the root directory instead of build directory
  - additionally update inline-scripted reference html files in the source test plan directory with `FileRecordChain`


jugglinmike:
Heads up: I won't be able to review this [until I can run the build script](https://github.com/w3c/aria-at/issues/890).

jugglinmike:
This branch now conflicts with the target branch. Because this has been open for a while, there's a possibility that the conflicts are substantial, so I'd prefer to wait on reviewing until after those have been addressed.

mzgoddard:
@jugglinmike I rebased onto master and fixed the conflicts and reproduced the changes on `process-test-directory.js`.

jugglinmike (commented):
The code looks good to me, and I've verified that its output is identical for the current "v1" and "v2" test plans.

That said, I spoke with Howard about the existing duplication in this file, and he had this to say:

> v1 csv resources and references eventually won't exist so there won't be a need to build them. A couple v1 files have already been removed since introducing v2
>
> So separating that into dedicated orchestrating scripts for the v1 and v2 test formats rather than creating branches in a single file that we may want to eventually prune in the future seemed much easier to do
>
> Some parts are similar enough though that the functions could be consolidated in a single util and used separately
>
> But a large majority of them felt like it would still adding a separate branch only to prune in the future from what I can remember and felt unnecessary, in that moment at least

Does that context impact that way you'd like to structure this patch?

howard-e (approved):
Looks good to me! Thanks for doing this work Z!

howard-e:
> ... I spoke with Howard about the existing duplication in this file, and he had this to say:
> 
> > v1 csv resources and references eventually won't exist so there won't be a need to build them. A couple v1 files have already been removed since introducing v2
> > So separating that into dedicated orchestrating scripts for the v1 and v2 test formats rather than creating branches in a single file that we may want to eventually prune in the future seemed much easier to do
> > Some parts are similar enough though that the functions could be consolidated in a single util and used separately
> > But a large majority of them felt like it would still adding a separate branch only to prune in the future from what I can remember and felt unnecessary, in that moment at least
> 
> Does that context impact that way you'd like to structure this patch?

@jugglinmike thanks for capturing my initial thoughts on this change!

Going back through this, I feel strongly now that the duplication can be significantly reduced. But it doesn't have to be done in this PR so as to unblock this. I've created #1089 to track that work.

PR #912, opened 2023-03-22
FileRecord.insert: merge directories and replace files

### LINKED ISSUE #910: Build process non-deterministically produces invalid HTML files
On my system (GNU/Linux, Ubuntu 18.04, Node.js 16.17.0), the `npm build` command occasionally outputs invalid HTML files.

    $ git show --oneline --no-patch
    b5fe3efd (HEAD -> master, upstream/master) Create tests for APG design pattern example: Horizontal Multi-Thumb Slider (#511)
    $ git status
    On branch master
    Your branch is ahead of 'bocoup/master' by 174 commits.
      (use "git push" to publish your local commits)
    
    nothing to commit, working tree clean

Steps to reproduce:

1. Build the HTML files five times, saving the output each time:

       $ for x in $(seq 5); do npm run build && mv build build-$x; done
2. Search the five build directories for files with multiple closing HTML tags:

       $ grep -Ro '</html' build-* | uniq -c | sort -n | grep -E '^\s*[2-9]'

Expected behavior: the search returns zero results

Actual behavior: the search returns one or more results

For my trial, the search returned:

    2 build-2/tests/command-button/reference/2022-4-8_124112/button.html:</html
    2 build-3/tests/tabs-manual-activation/reference/2021-2-15_124757/tabs.html:</html
    2 build-5/tests/form/reference/2021-10-24_152221/form.html:</html
    2 build-5/tests/modal-dialog/reference/2022-4-7_15544/dialog.html:</html

The symptom is more severe than duplicated closing tags; that just happens to be a straightforward way to detect the problem.

<details>
  <summary>Here's a comparison highlighting the invalidity in one of those files</summary>

```diff
$ git diff -w --no-index build-1/tests/command-button/reference/2022-4-8_124112/button.html build-2/tests/command-button/reference/2022-4-8_124112/button.html
diff --git a/build-1/tests/command-button/reference/2022-4-8_124112/button.html b/build-2/tests/command-button/reference/2022-4-8_124112/button.html
index d17cbeb1..8d7a8451 100644
--- a/build-1/tests/command-button/reference/2022-4-8_124112/button.html
+++ b/build-2/tests/command-button/reference/2022-4-8_124112/button.html
@@ -7,23 +7,7 @@
   <!-- CSS and JS for this example. -->
   <link rel="stylesheet" href="css/button.css">
   <script src="js/button.js"></script>
-
-<!-- Generated by create-example-tests.js -->
-<script>
-  (function() {
-    function setupScript(testPageDocument) {
-      //
-
-    };
-    document.addEventListener('click', function(event) {
-      if (event.target.classList.contains('button-run-test-setup')) {
-        event.target.disabled = true;
-        setupScript(document);
-      }
-    });
-  })();
-</script>
-<!-- End of generated output --></head>
+</head>^M
 <body>
   <main>
     <h1>Command Button Example</h1>
@@ -48,6 +32,18 @@
         <a id="afterlink" href="#afterlink">Navigate backwards from here</a>
       </div>
 
+      <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>^M
+    </section>^M
+  </main>^M
+</body>^M
+</html>^M
+s from here</a>
+        <div>
+          <div tabindex="0" role="button" id="action">Print Page</div>
+        </div>
+        <a id="afterlink" href="#afterlink">Navigate backwards from here</a>
+      </div>
+
       <div role="separator" id="ex_end_sep" aria-labelledby="ex_end_sep ex_label" aria-label="End of"></div>
     </section>
```

</details>

I've scraped all of the test HTML which is currently published on the production deployment of ARIA-AT App, and the good news is that  it appears to be valid.

That may mean that the most recent build was lucky (notice how in my trial, the first and fourth build didn't include any duplicated closing HTML tags), or it may mean that this problem is specific to my machine. I'm wondering if anyone else can reproduce the behavior reported here.

mzgoddard:
[Preview Tests](https://deploy-preview-912--aria-at.netlify.app)

Fix #910 

When inserting directories, walk its entries and merge items into the record. When inserting files, replace buffer with new buffer.

-----

This is a change to FileRecord.insert that I think follows how the method currently works for the path parameter to the record parameter. Before this change inserting a directory or file record would append it to the parent directory stated by the path parameter. That causes some odd outcomes when adding items that are already in the record.

Seeing #910 I think this is a reasonable change to make FileRecord work how I think others would anticipate. An alternative we could do would be to make FileRecord more strict and instead of merging and replacing records, error if there is a record already at the given path.

### DIFF
--- a/lib/util/file-record.js
+++ b/lib/util/file-record.js
@@ -379,16 +379,24 @@ function createHost({ fs = _fs.promises, path = _path } = {}) {
       record.entries = [];
     }
 
-    if (splitPath.length > 1) {
-      let entry = record.entries.find(({ name }) => name === splitPath[0]);
-      if (!entry) {
-        entry = { name: splitPath[0] };
-        record.entries.push(entry);
-      }
+    const name = splitPath[0];
+    let entry = record.entries.find(subrecord => subrecord.name === name);
+    if (!entry) {
+      entry = { name };
+      record.entries.push(entry);
+    }
 
+    if (splitPath.length > 1) {
       insertInRecord(entry, splitPath.slice(1), clone(item));
-    } else {
-      record.entries.push({ name: splitPath[0], ...clone(item) });
+    } else if (isDirectory(item)) {
+      for (const subentry of item.entries) {
+        insertInRecord(entry, [subentry.name], subentry);
+      }
+    } else if (isFile(item)) {
+      if (isDirectory(entry)) {
+        throw new Error('Cannot insert file where a directory exists.');
+      }
+      entry.buffer = item.buffer;
     }
   }
 


jugglinmike (approved):
Works for me!

PR #986, opened 2023-08-24
menu-button-actions-active-descendant: test different menu invocation approaches in setup scripts

jscholes:


jscholes:
@howard-e The "Preview Tests" link doesn't seem to work; Netlify says that the site is not found.  Could you please take a look?

howard-e:
> @howard-e The "Preview Tests" link doesn't seem to work; Netlify says that the site is not found. Could you please take a look?

@jscholes navigating to [the Preview Tests link](https://deploy-preview-986--aria-at.netlify.app/) should work as expected now.

Netlify "lost" the auth connection to this repository. So I've relinked the repo. This has happened in some other repositories earlier this year. It's been discussed in some Netlify forums and still unsure why it happens, but relinking the repo works.

Thanks for raising this!

PR #991, opened 2023-09-28
Refactor Link Example 1 (span element with text content) test plan to V2 test format

IsaDC:
[Preview Tests](https://deploy-preview-991--aria-at.netlify.app)


mcking65 (commented):
@IsaDC 

I made one correction to the nav back to link command for VoiceOver. I change it to "shift+l". It was previously "l".

After doing that, I noticed there are 32 files changed in this PR. In addition to the link tests, there are changes to button and meter as well as to package.json and package-lock.json.


The easiest way to fix that is make a fresh PR with the files we want. I'll close this PR and make that new one, including all the current link/data.

mcking65:
replaced by #1013

PR #992, opened 2023-10-02
Refactor Command Button test plan to V2 test format

IsaDC:
[Preview Tests](https://deploy-preview-992--aria-at.netlify.app)


mcking65 (approved):
Thanks @IsaDC , looks great. I made only 2 changes:
1. Removed out-dated files.
2. Corrected one error in VO where nav back had j instead of shift+j

PR #990, opened 2023-09-25
V2-maker Changes

howard-e:
* Removes whitespace in `support.json`
* Replaces instance of commands that had 'one' or 'two' with their numerical values as `shift+one` or `two` isn't a command that can be translated with the new `commands.json`

Issue #977, opened 2023-08-14
Update test builder to support V2 of test format


mcking65:
#### Objective

Revise the [aria-at build scripts](https://github.com/w3c/aria-at/tree/master/scripts) so they support not only V1 but also the new V2 test format defined in #974.

While versions of the test plans that are written in the V1 format will all eventually be deprecated, for at least a few years, we will continue to need the ability to view plans written in the V1 format.

Note: Consider moving the directory `tests/resources` to `scripts/resources` or some other location outside of `tests` since the content of that folder is not related to any specific test plan and appears to be related to building the test plans.

#### Changes to build review files

When running the build, a review file for each test plan is generated in the directory `build/review`. When the builder is building a test plan that is written in the V2 format, it should generate an HTML file that has the following differences when compared to the files currently being generated for the V1 format.

This zip file contains a mockup of a portion of the slider test plan in V2 format.
[horizontal-slider.zip](https://github.com/w3c/aria-at/files/12339730/horizontal-slider.zip)

Note that this description is referring to properties in support.json that are visible in the slider-refactor branch via #975. That branch includes a refactor of the horizontal-slider test plan into the V2 format.

#### Page title

Change the H1 to have format `TITLE Test Plan (NUMBER_OF_TESTS tests)` where TITLE is the value of the `title` refId in `references.csv`.:

Example:
```
<h1>Color Viewer Test plan (9 tests)</h1>
``````

#### Filter radio group

Change the label of the radio group to communicate that the radios represent the AT covered by the plan:

Example:

```
<legend>Filter tests by covered assistive technology</legend>
```

To ensure there is a radio button for each AT that is covered, use the names of"AT_KEY-command.csv" files as the source of which AT are covered. For each command CSV file, include a radio button labeled using the corresponding `name` property in `support.json`.

#### Supporting documentation section

After the filters, add a level 2 heading with text "Supporting Documentation".

After the heading, list any of the  following references that are included in `references.csv`. Since these apply to the entire test plan, do not include these references in the list of relevant specifications under the `h2` for each test:
* example
* designPattern
* developmentDocumentation

For each of the above `refIds`, include a link using link text from the `linkText` column and the URI from the `value` column.

Example:
```
<ul>
    <li><a href="https://www.w3.org/WAI/ARIA/apg/patterns/slider/examples/slider-color-viewer/"">APG Color Viewer Slider Example</a></li>
[...]

PR #997, opened 2023-10-12
Support V2 Test Format build

howard-e:
[Preview Tests](https://deploy-preview-997--aria-at.netlify.app)

This PR adds support for building review pages for the new [v2 Test Format](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2), along with supporting more specific requirements detailed in #977 (for making changes to the review page, not the test page).

Any work in this PR related to the v2maker files is included in #990 and should be reviewed separately there.

I've also submitted #998 which makes the v2 file formats of all the tests. The [link is here](https://deploy-preview-998--aria-at.netlify.app).

Newly supported example feature changes to [assertion priorities](https://github.com/w3c/aria-at/blob/237cd6732fe7188fea9e29d992c4c5cbca638419/tests/alert/data/jaws-commands.csv), [tokenized assertions and refs](https://github.com/w3c/aria-at/blob/237cd6732fe7188fea9e29d992c4c5cbca638419/tests/alert/data/assertions.csv) have been set for [998's alert](https://deploy-preview-998--aria-at.netlify.app/review/alert).

**Note:** The [V1 Test Format](https://github.com/w3c/aria-at/wiki/Test-Format-V1-Definition) is still supported.

Change to the scripts:
* `npm run build` - This script will check to see if relevant v2 test format files exist to build the relevant pages first, otherwise it falls back to building with the v1 test format files.
* `npm run build-v1` - Explicitly builds the review pages using the v1 test format in verbose mode.
* `npm run build-v2` - Explicitly builds the review pages using the v2 test format in verbose mode (this assumes that valid v2 test format files exist in a pattern's directory; this can be created with `npm run make-v2 {pattern}` if it doesn't exist. #998 has created the files created for all the directories). 

Known enhancements which can be a part of this PR or added in a follow-up:
* Handling space-separated settings ~but the use case it aims to solve can already be supported. Is this requirement still needed?~ (See #1002)
* ~Capitalize single length instances of commands (letters). Capitalizing the values in commands.json may be best.~ ([not doing](https://github.com/w3c/aria-at/pull/997#issuecomment-1804637348))
* Move `<root>/tests/resources` to `<root>/resources`.
* Update the individual test page to also account for 0-level assertion exceptions.


stalgiag (commented):
I was able to do a first round of changes. This PR constitutes a substantial body of work and will need more testing. I am just adding documentation of my first pass on testing and confirming behavior against the outline provided in #977 

#### Changes to Build Review Files

**Page Title**
- [x] Update H1 to follow new format: `TITLE Test Plan (NUMBER_OF_TESTS tests)`.

**Filter Radio Group**
- [x] Change label of radio group.
- [x] Create a radio button for each covered AT, sourced from "AT_KEY-command.csv" and labeled from `support.json`.

**Supporting Documentation Section**
- [x] Add an H2 heading with the text "Supporting Documentation".
- [x] Include references listed in `references.csv` that apply to the entire test plan.

**Changes in List Under Test Title**
- [x] Remove screen reader mode specification.
- [x] Remove "applies to" statement.
- [x] Add a preface paragraph from `ariaSpecsPreface`. (Unable to test but discussed with Howard and present in code)
- [x] Modify the list items' content to include links for refIds from `assertions.csv`. (Unable to test but discussed with Howard and present in code)

##### Changes under H3 for Each AT

- [x] Remove H4 and content covering scripted instructions.
  
**Tester Instructions**
- [x] Rename H4 to "Instructions".
- [x] Insert default configuration as the first list item.
- [x] Remove instructions about screen reader mode.
- [x] Modify subsequent list items with task instructions and commands from `AT_KEY-commands.csv`.

**Command-specification string and settings**
- [x] Add paragraphs and ordered lists to explain any specified settings.

**Assertions Table**
- [ ] Replace the current table with tables sorted by commands. (TODO: confirm is sorted by commands)
- [x] Add tables for each command with columns for Priority, Assertion Phrase, and Assertion Statement. Sort them by priority and assertion sequence.

mcking65:
@howard-e 

I am seeing some consistent problems across all test plans with the preview. For example, consider [Color Viewer Slider Test Plan | For Pattern: horizontal-slider](https://deploy-preview-997--aria-at.netlify.app/review/horizontal-slider).

The preview:
1. Shows 21 tests. But there are only 9 tests in tests.csv.
2. Shows different tests for VoiceOver vs JAWS and NVDA.
3. Does not show the settings in the command headings e.g., Tab with Virtual Cursor Active
4. Because settings are ignored, some commands are missing . the test for navigate forward should include Tab with PC cursor active and Tab with virtual cursor active, but the navigation tests only have one Tab command.
5. The assertion phrase is shown as "N/A" instead of showing the actual assertion phrase
There must be some bugs in #990 because it produces different output when compared to v2maker in master. It is using the wrong rows in tests-v1.csv. This is causing the test titles to include modes and not properly grouping commands in the command files. I think the bugs in #990 are one of the reasons for some of the above problems. One way to test this hypothesis would be to use the output from the version of v2maker in master to see which problems go away.

howard-e:
> I am seeing some consistent problems across all test plans with the preview. For example, consider [Color Viewer Slider Test Plan | For Pattern: horizontal-slider](https://deploy-preview-997--aria-at.netlify.app/review/horizontal-slider).

Before continuing (and because I haven't gotten the chance to fully share out the full details of the updates), there are 2 views being represented here.

This PR is using only the v1 format of the files against the newly described content of #977. #998 is a supporting to PR to show what the content looks like with with the actual v2 format created by the make-v2 script. I'd rather not include those v2 files in this PR, so we can leave that to the test authors as they recreate these files for each test pattern (and make updates where necessary), and to also show feature parity since we're supporting both v1 and v2 for now.

With that in mind, could you re-confirm your follow up points after going through https://deploy-preview-997--aria-at.netlify.app/review/horizontal-slider instead?

> 1. Shows 21 tests. But there are only 9 tests in tests.csv.

This seems consistent with what is currently displayed at https://aria-at.netlify.app/review/horizontal-slider. What is the expectation?

mcking65:
@howard-e The number of tests is now correct in the V2 previews. The presentation of screen readers and commands under each screen reader is now correct in the V2 format.

There are a couple of strings that should be adjusted; I need to check if they are coming from the JSON. James and I discussed using details/summary elements for the settings-related instructions. I know I prototyped that, but I don't know if I specified it. I will check on that soon. These are really just cosmetic issues that will have no impact on the downstream work on the other P0 requirements in the refactor project.

Regarding the following notes you wrote about unfinished work ...

>Capitalize single length instances of commands (letters). Capitalizing the values in commands.json may be best. 

We are not going to capitalize them. The command "b" is not the same as the command "B". The single-letter commands are case-sensitive. "Shift+b" is the same as "B", and it is more clear to specify "Shift+b". Apple does this in its settings. If you look at the VoiceOver settings for Quick Nav, they write "B" and "b" as two different commands. We plan to specify "b" and "Shift+b" to avoid any ambiguity. Similarly, we will specify "1" and "Shift+1", not "1" and "!".

>Update the individual test page to also account for 0-level assertion exceptions. 

I am not sure I understand what you mean by "individual test page". Is "0:Assertion_ID" not yet supported if it is specified in AT_NAME-commands.csv?

howard-e:
> > Capitalize single length instances of commands (letters). Capitalizing the values in commands.json may be best.
> 
> We are not going to capitalize them. The command "b" is not the same as the command "B". The single-letter commands are case-sensitive. "Shift+b" is the same as "B", and it is more clear to specify "Shift+b". Apple does this in its settings. If you look at the VoiceOver settings for Quick Nav, they write "B" and "b" as two different commands. We plan to specify "b" and "Shift+b" to avoid any ambiguity. Similarly, we will specify "1" and "Shift+1", not "1" and "!".

Thanks for that context!

> > Update the individual test page to also account for 0-level assertion exceptions.
> 
> I am not sure I understand what you mean by "individual test page". Is "0:Assertion_ID" not yet supported if it is specified in AT_NAME-commands.csv?

It is supported on the review page ([eg. Test 1: Trigger an alert > JAWS > Enter (virtual cursor active) > does not show the assertion `'MAY' priority, 'convey some phrase'`](https://deploy-preview-998--aria-at.netlify.app/review/alert), that was [included as a test](https://github.com/w3c/aria-at/blob/build-v2-test-format-with-new-csvs/tests/alert/data/jaws-commands.csv)) but is present for the other commands.

On the [JAWS collection page](https://deploy-preview-998--aria-at.netlify.app/tests/alert/test-03-triggeralert?at=jaws), it is still shown under `Enter (virtual cursor active)`. I'd think we don't want to display it there as well.

mcking65 (commented):
Looking good now. We should merge this ASAP to unblock reviews of the V2 tests.

However, we need to fast follow with two additional pull requests.

### Change 1: Support for 0-priority assertions in the test collection form

It is critical 0-priority aassertions do not show up in the collection form. Example, the assertion

>JAWS switched from virtual cursor active to PC cursor active

in [JAWS Trigger an alert](https://deploy-preview-998--aria-at.netlify.app/tests/alert/test-03-triggeralert?at=jaws).

### Change 2: Remove Success Criteria section from collection form

Similar to the reports and single page view of the test plan in the app, we need to remove the following content from the collection form:

>## Success Criteria
>
>To pass this test, JAWS needs to meet all the following assertions when each specified command is executed:
>
>1. Role 'alert' is conveyed 
>2. Text 'Hello' is conveyed 
>3. JAWS switched from virtual cursor active to PC cursor active 

We have removed the concept of passing a complete test; we pass/fail individual assertions but at the level of a test, we simply track the number/percentage of passing and failing assertions of each priority, so this content is no longer relevant.

Issue #969, opened 2023-08-04
Change results collection form to support pass/fail assertion verdicts


mcking65:
This issue is one element of the plan documented in #961 to migrate from 3 assertion verdicts (good/no output/incorrect output) to two verdicts (pass/fail).

In the results collection form, there is a three-column table of assertions and radio buttons for recording the verdicts of each assertion. Replace the table with a `fieldset` containing checkbox elements with assertion statements as labels.

This specification assumes this change is being made coincident with or after w3c/aria-at-app#743. Thus, it specifies [Test Format V2](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2) values.

The `legend` of the `fieldset` is composed of the following strings:

`assertionResponseQuestion + " " + commandSpecificationString + "?"`

Where:
* `assertionResponseQuestion` is the `name` property of an object in the `testPlanStrings` array in `support.json`
* `commandSpecificationString` is a command or command sequence string with any associated settings in parentheses as described in #977.

Using values from `support.json` in the `slider-refactor` branch and commands and assertions from the `jaws-commands.csv` in the same branch, here is example code:

~~~
<fieldset>
  <legend>Which statements are true about the response to Down Arrow then Down Arrow (virtual cursor active)?</legend>
  <label>
    <input type="checkbox" name="assertion1">
    Role 'slider' is conveyed
  </label>
  <label>
    <input type="checkbox" name="assertion2">
    Name 'Red' is conveyed
  </label>
  <label>
    <input type="checkbox" name="assertion3">
    Value '128' is conveyed
  </label>
  <label>
    <input type="checkbox" name="assertion4">
    Orientation 'horizontal' is conveyed
  </label>
  <label>
    <input type="checkbox" name="assertion5">
    Minimum value '0' is conveyed
  </label>
  <label>
    <input type="checkbox" name="assertion6">
    Maximum value '255' is conveyed
  </label>
</fieldset>
~~~

mcking65:
@jscholes We will discuss at Tuesday meeting; please look at attached HTML before then if possible.

jscholes:
@mcking65 I feel very strongly that radios are not an optimal direction, and also that the assertions tables for testers are not a helpful construct to keep going forward.  I've expanded on my reasoning in https://github.com/w3c/aria-at/issues/974#issuecomment-1676029956, but let's discuss more this coming Tuesday if possible.

mcking65:
@jscholes 

[...]

PR #1003, opened 2023-11-08
Update results collection form to support pass/fail assertion verdicts

stalgiag:
[Preview Tests](https://deploy-preview-1003--aria-at.netlify.app)

see #969 
blocked from merging by #997 

This is an initial pass. These files are new to me so there is likely something I am missing (especially with output/serialization). I will leave this is as a draft for now and will have some side-channel conversations with others to ensure I am addressing everything.

Also, note that this is currently targeting `build-v2-test-format` to make viewing the diffs easier but this will target `main` once that branch is merged.

### DIFF
--- a/lib/data/process-test-directory.js
+++ b/lib/data/process-test-directory.js
@@ -528,6 +528,7 @@ const processTestDirectory = async ({ directory, args = {} }) => {
       setup_script_description: getSetupScriptDescription(test.setupScript.scriptDescription),
       setupTestPage: test.setupScript.script,
       specific_user_instruction: test.instructions,
+      assertionResponseQuestion: supportJson.testPlanStrings.assertionResponseQuestion,
       commandsInfo: test.commandsInfo,
       output_assertions: test.assertions,
     };
--- a/tests/resources/aria-at-test-run.mjs
+++ b/tests/resources/aria-at-test-run.mjs
@@ -243,9 +243,9 @@ export function instructionDocument(resultState, hooks) {
         change: atOutput => hooks.setCommandOutput({ commandIndex, atOutput }),
       },
       assertionsHeader: {
-        descriptionHeader: `${
-          resultState.assertionResponseQuestion ?? 'Which statements are true about the response to'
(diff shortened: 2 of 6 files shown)


lib/data/process-test-directory.js
```diff
@@ -528,6 +528,7 @@ const processTestDirectory = async ({ directory, args = {} }) => {
       setup_script_description: getSetupScriptDescription(test.setupScript.scriptDescription),
       setupTestPage: test.setupScript.script,
       specific_user_instruction: test.instructions,
+      assertionResponseQuestion: supportJson.testPlanStrings.assertionResponseQuestion,
```
howard-e, line 531: You should also include the same in `process-test-directory-v1`, since file also has references to `support.json`. And we'd expect to see the same string in the v1 output if `support.json` changes.

tests/resources/aria-at-test-run.mjs
```diff
@@ -261,9 +243,9 @@ export function instructionDocument(resultState, hooks) {
         change: atOutput => hooks.setCommandOutput({ commandIndex, atOutput }),
       },
       assertionsHeader: {
-        descriptionHeader: 'Assertion',
-        passHeader: 'Success case',
-        failHeader: 'Failure cases',
+        descriptionHeader: `${
+          resultState.assertionResponseQuestion ?? 'Which statements are true about the response to'
```
howard-e, line 247: I'm guessing this nullish check was for it missing, when building against v1. But my previous comment should address not needing it, if you'd like to remove it here. It's not unlike this section to have default hardcoded fallbacks so I don't feel strongly on removing it or not.

howard-e (changes requested):
@stalgiag looks good! I've verified it works as expected across both the v1 and v2 builds, and having the output like this makes things a lot easier.

I've left a request for minor change, but this should be good to merge whenever #977 can be merged.

tests/resources/aria-at-test-run.mjs
stalgiag: Removed. Thanks!

howard-e (approved):
@stalgiag thanks for addressing the feedback. LGTM!

PR #1005, opened 2023-11-21
Include content from support.json properties when using `loadCollectedTestAsync`

howard-e:
[Preview Tests](https://deploy-preview-1005--aria-at.netlify.app)

To support missing `support.json` properties being included in `*.collected.html`.


howard-e:
Superseded by #1017

PR #1006, opened 2023-11-29
Update collection form to capture severity and additional details on an unexpected behavior happening

howard-e:
[Preview Tests](https://deploy-preview-1006--aria-at.netlify.app)

This is additional work to support https://github.com/w3c/aria-at-app/issues/738.

> Modify the results collection form so that each behavior in the list of problematic behaviors has the following inputs:
> - A checkbox to indicate if the behavior occurred.
> - A select labeled "Impact" with values Moderate, and High. It is disabled if the checkbox is not checked and required if it is checked.
> - A text field labeled "Details" that is disabled if the checkbox is not checked and required if it is checked.


howard-e:
Superseded by #1016

PR #1007, opened 2023-11-30
Fix for `.collected.html` failing to render for v1

howard-e:
[Preview Tests](https://deploy-preview-1007--aria-at.netlify.app)

Fix issue with `commands.json (v2)` failing to load when building v1 with clean build.

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -114,19 +114,17 @@ export function verifyATBehavior(atBehavior) {
 }
 
 export async function loadCollectedTestAsync(testRoot, testFileName) {
-  let commandsJsonResponse;
-  let commandsJson;
+  const collectedTestResponse = await fetch(`${testRoot}/${testFileName}`);
+  const collectedTestJson = await collectedTestResponse.json();
+
   try {
     // v2 commands.json
-    commandsJsonResponse = await fetch(`../commands.json`);
-    commandsJson = await commandsJsonResponse.json();
+    const commandsJsonResponse = await fetch(`../commands.json`);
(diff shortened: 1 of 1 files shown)


tests/resources/aria-at-harness.mjs
```diff
@@ -114,12 +114,19 @@ export function verifyATBehavior(atBehavior) {
 }
 
 export async function loadCollectedTestAsync(testRoot, testFileName) {
-  const commandsJsonResponse = await fetch(`../commands.json`);
-  const commandsJson = await commandsJsonResponse.json();
+  let commandsJsonResponse;
```
gnarf, line 117: Could probably just be const inside the try now - it's not used outside

tests/resources/aria-at-harness.mjs
```diff
@@ -114,12 +114,19 @@ export function verifyATBehavior(atBehavior) {
...
+    commandsJsonResponse = await fetch(`../commands.json`);
+    commandsJson = await commandsJsonResponse.json();
+  } catch (e) {
+    // v2 commands.json isn't available
+  }
 
   const collectedTestResponse = await fetch(`${testRoot}/${testFileName}`);
   const collectedTestJson = await collectedTestResponse.json();
-  testRunIO.setAllCommandsInputFromJSON(commandsJson);
+  if (commandsJson) testRunIO.setAllCommandsInputFromJSON(commandsJson);
```
gnarf, line 129: any reason not to just fold this up into the the try as well, get rid of the the let scope entirely?

gnarf (approved):
This will definitely work, couple of random suggestion/questions added but feel free to land as is if needed.

PR #1011, opened 2023-12-05
Update process-test-directory.js assertionExceptions regex

howard-e:
[Preview Tests](https://deploy-preview-1011--aria-at.netlify.app)

Update process-test-directory.js to support building assertionExceptions pointing to assertionIds ending with numeric value.


howard-e:
Included in #1012

Issue #1008, opened 2023-12-03
Do not show 0-priority assertions in the results collection form


mcking65:
Add support for 0-priority assertions in the test collection form. This means that they should not be included in the collection form.

Example, the assertion

>JAWS switched from virtual cursor active to PC cursor active

in [JAWS Trigger an alert](https://deploy-preview-998--aria-at.netlify.app/tests/alert/test-03-triggeralert?at=jaws).

_Originally posted by @mcking65 in https://github.com/w3c/aria-at/pull/997#pullrequestreview-1753673636_

mcking65:
I have verified that previews of unmerged test plans with 0 priority assertions work correctly. There are no merged test plans with that characteristic, so I don't know how to verify the runner presentation.

PR #1012, opened 2023-12-05
Exclude 0-level priority assertions in the collection form

howard-e:
[Preview Tests](https://deploy-preview-1012--aria-at.netlify.app)

See #1008.

To test this change:
* Run `npm run make-v2 <pattern>`
* Update any of the created `*-commands.csv`'s row value for the _"assertionExceptions"_ column with `0:<assertionId> ...`
  * Note: `assertionId` must match a defined assertionId of a row in the generated `assertions.csv`
  * Note: `tests.csv` must include the assertion id in a row's _"assertions"_ column
* Run `npm run build`
* Observe that the assertion is excluded on `<baseurl>/review/<pattern>.html` under the expected command (this was included before this PR)
* Select an AT link under tests and observe that the assertion is excluded on `<baseurl>/tests/<pattern>/<testfile_name>.html` under the expected command

**Note that #1014 has also been submitted to make previewing those changes easier (check description for changes and [preview link](https://deploy-preview-1014--aria-at.netlify.app/))**


alflennik (approved):
I was able to verify that excluded assertions do not appear. Code looks good.

stalgiag (approved):
Tested and I was able to exclude assertions. I didn't notice any issues with the code.

PR #1016, opened 2023-12-07
Update collection form to capture unexpected behaviors severity and additional details

howard-e:
Supersedes #1006

This is additional work to support https://github.com/w3c/aria-at-app/issues/738.

> Modify the results collection form so that each behavior in the list of problematic behaviors has the following inputs:
> - A checkbox to indicate if the behavior occurred.
> - A select labeled "Impact" with values Moderate, and High. It is disabled if the checkbox is not checked and required if it is checked.
> - A text field labeled "Details" that is disabled if the checkbox is not checked and required if it is checked.

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -254,6 +254,8 @@ const h3 = bind(element, 'h3');
 const hr = bind(element, 'hr');
 const input = bind(element, 'input');
 const label = bind(element, 'label');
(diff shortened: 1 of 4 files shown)


alflennik (approved):
Looks great @howard-e, I was able to add unexpected behaviors with the new UI structure including the severity and details, and see the unexpected behaviors listed in the results. I also verified that aria-at-app and aria-at have equivalent behavior.

PR #1020, opened 2023-12-13
support.json: Add VO settings for new single and arrow quick key nav added to VO in latest macOS

mcking65:
[Preview Tests](https://deploy-preview-1020--aria-at.netlify.app)

In a recent build of macOS, quick nav was split into two separate settings: one for single letter keys and one for arrow keys. Now the settings required for a test depends on whether the command sequence being tested is an arrow key command, quick nav letter command, or both.


mcking65:
I am taking the chance of merging without further review. We have to mod some tests in a hurry. If I messed up, I'll fix with another PR. This should be low risk.

Issue #1039, opened 2024-02-20
Enhance V2 test format to allow 0 priority assertions in tests.csv


mcking65:
Currently, if a test includes an assertion that applies to only one of several commands, the assertion is listed in the assertions column of tests.csv and then it is removed from commands where it is not wanted by specifying a priority of 0 in the assertionsExceptions column of the `*-commands.csv` files. This creates the need for lots of exceptions when there is an assertion that is only applicable to one or two commands out of several and only applies to a subset of the in scope AT.

This proposal aims to simplify the csv file content by inverting the logic for such scenarios. If we could specify an assertion priority of 0 in tests.csv, that would make it possible to override the 0 priority with 1, 2, or 3 for only the commands where that assertion is applicable.

One might ask why is it necessary to specify the assertion in tests.csv at all. Why not simply add it where desired in the command.csv files. The answer is that the order of assertions in the presentation of assertions for a command is determined by the order of the assertionIDS in the assertions column of tests.csv. We could avoid the need to include the assertion in tests.csv in scenarios where it is desired for the exceptional additional assertion to be the last assertion in the presentation of the assertions for a command. That is, we could specify that if an assertion is not included in tests.csv but is included in commands.csv for a given test, then that assertion is appended to the list generated from tests.csv. If such logic is not too opaque, that would be a reasonable way to further simplify the csv files in some scenarios.

I've discovered there are also some scenarios where it would be helpful to replace a specific assertion with a different one for only some commands. For example, when navigating backward into a radio group where the first button in the group is checked, using shift+tab in reading mode moves to the first radio where as using up arrow moves the reading cursor to the last radio. This is expected. In this case, it would be helpful for the tests.csv to include assertions for both the first and last radio but with different priorities so only one is included. Then, in commands.csv, the shift+tab command could have exception that turn off assertions related to the third radio and turn on assertions related to the first.

In PR #1040, tests.csv and commands.csv illustrate these scenarios.

mcking65:
@ccanash  @howard-e 

Please size and prioritize this ASAP. It provides a far superior path forward for the radio group test plan and test plans for any other similar composite widgets that require testing of mode switching and have similar challenges where shift+tab differs from other commands when navigating backward.

ccanash:
Hi @mcking65  confirming this has entered our triaging process

howard-e:
#1042 should address this

mcking65:
[...]

PR #1040, opened 2024-02-20
[DO NOT MERGE] Test plan that illustrates assertion exception enhancement specified in issue 1039

mcking65:
[Preview Tests](https://deploy-preview-1040--aria-at.netlify.app)

Issue #1039 proposes an enhancement to the logic for assertion exceptions. The first two tests in the radio group test plan in this PR illustrates how that logic would be used.

In the tests.csv file, the assertions for the first test, `navForwardsRadioGroupWhereNoRadioButtonsAreChecked`, includes `0:interactionModeEnabled`. This specifies that the assertion `interactionModeEnabled` is not included for any commands associated with this test unless a specific command includes an assertion exception overrides the 0 priority. In jaws-commands.csv and nvda-commands.csv, there is one command associated with this test, `tab` with a readingCursor setting, that overrides the priority 0 with priority 2.

Similarly, in the test `NavBackIntoRadioGroupWhereLastRadioUnchecked`, the assertions column includes:

`nameThinCrust 0:nameRegularCrust`

So, by default, the `nameThinCrust` assertion is included for all commands. However, for the shift+tab command, each of the screen reader command files includes the assertionException of:

`0:nameThinCrust 1:nameRegularCrust`

This is because when navigating backward with shift+tab, the cursor moves to the regular curst radio instead of the thin crust radio.

As a result, if #1039 is implemented, the tab and shift+tab commands for these two tests would have assertions that are different from the assertions for other commands.


howard-e:
#1042 should now make this possible.

mcking65:
Purposed served; #1042 is merged. closing

PR #1049, opened 2024-03-26
Refactor Modal Dialog test plan to V2 test format (update)

IsaDC:
[Preview Tests](https://deploy-preview-1049--aria-at.netlify.app)

- Run make-v2 utility to convert the Modal Dialog test plan to the V2 test format
- Update test instructions to V2 format


IsaDC:
@mcking65  This test plan is ready for review.
Thank you.

mcking65:
Questions to discuss:

1. When closing a modal with JAWS and pc cursor is active, if user pressed insert+z, only one press of escape is needed. What should we assume about the state of JAWS?
2. Could we combine the tests:
    * Test 3: Close a modal dialog using a button
    * Test 2: Close a modal dialog
    * Or, if not combine, make the actions more distinct, e.g. Escape from a modal and close a modal
3.  in Test 4: Navigate to the last focusable element in a modal dialog and Test 5: Navigate to the first focusable element in a modal dialog, shouldn't the commands also include JAWS VC and NVDA Browse commands?
4. While testing navigation to abeginning and end of the modal ensures the screen reader recognizes the boundaries, it doesn't test whether the screen reader enforces the boundaries. Should we have a test that is something like "Try to navigate beyond boundaries of a modal dialog"?

IsaDC:
> Questions to discuss:
> 
> 1. When closing a modal with JAWS and pc cursor is active, if user pressed insert+z, only one press of escape is needed. What should we assume about the state of JAWS?
> 2. Could we combine the tests:
>    
>    * Test 3: Close a modal dialog using a button
>    * Test 2: Close a modal dialog
>    * Or, if not combine, make the actions more distinct, e.g. Escape from a modal and close a modal
> 3. in Test 4: Navigate to the last focusable element in a modal dialog and Test 5: Navigate to the first focusable element in a modal dialog, shouldn't the commands also include JAWS VC and NVDA Browse commands?
Matt,
I've implemented these changes.

> 4. While testing navigation to the beginning and end of the modal ensures the screen reader recognizes the boundaries, it doesn't test whether the screen reader enforces the boundaries. Should we have a test that is something like "Try to navigate beyond boundaries of a modal dialog"?
I have not prepared a draft yet, as I would like to discuss some questions during our next meeting.

mcking65:
@IsaDC 

I think we can remove the test:

>Test 10: Open a nested modal dialog using a link

The only purpose of that link is so that there is more than one focusable element in the dialog. Links should open pages, not dialogs; it only opens a dialog in that example because we didn't want the user to leave the example page.

mcking65:
@IsaDC 

I think we could combine the tests:

>Test 8: Close a nested modal dialog
>
>Test 9: Close a nested modal dialog using a button

In the same way that you combined the other close dialog tests.

Issue #1046, opened 2024-03-11
Support test plans that specify a command assertion exception that does not include an assertion priority prefix


mcking65:
This is a test writing simplification proposal that is related to #1045. It might be a good idea to implement both #1045  and this request in a single PR.

Currently, assertion exceptions in the `*-commands.csv` files are required to have a priority prefix of `n:` where `n` is 0, 1, 2, or 3. This proposal is to support omitting the priority prefix in the `assertionExceptions` column of `*-commands.csv` in situations where the test author wants the priority specified in `tests.csv` or `assertions.csv` to be assigned to the assertion.

For example, the priority of assertions for switching screen reader mode is specified in assertions.csv. It is typically `2`. It is redundant to require the test author to specify it again in the commands file. Requiring the author to repeat the priority in the commands file creates opportunity for accidental inconsistency across AT or even among different commands for a specific AT.

The current priority precedence defined in the V2 format would not change. That is:
1. The priority specified in commands.csv is assigned if specified.
2. If priority is not specified in commands.csv, the priority specified in tests.csv is assigned.
3. If priority is not specified in tests.csv, the priority specified in assertions.csv is assigned.

ccanash:
Hi @mcking65 can you please confirm this works as expected?  Please see comment in #1053

PR #1053, opened 2024-04-24
Support assertionExceptions without priority prefix

evmiguel:
This PR addresses #1046, providing support for test plans with assertionExceptions that do not have a priority prefix.

The following changes were added:

- Relaxed regex check on assertionExceptions
- Util file for setting default assertion priorities

### DIFF
--- a/lib/data/parse-test-csv-row.js
+++ b/lib/data/parse-test-csv-row.js
@@ -3,6 +3,8 @@
 
 'use strict';
 
+const { setDefaultAssertionExceptionsFromTest } = require('../util/assertion-exception');
+
 /**
  * @param {AriaATCSV.Test} testRow
  * @returns {AriaATParsed.Test}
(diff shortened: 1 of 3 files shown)


lib/data/parse-test-csv-row.js
```diff
@@ -216,12 +218,16 @@ function parseTestCSVRowV2({ tests, assertions, scripts, commands }) {
             at => at.key === key && at.settings === (command.settings || 'defaultMode')
           )
         ) {
+          const assertionExceptions = setDefaultAssertionExceptionsFromTest(
```
howard-e, line 221: `setDefaultAssertionPriority` seems like a more appropriate name here

howard-e (approved):
@evmiguel looks good! Well done, in also considering a lot of these files are new to you.

How this is done also makes it so that there shouldn't be any additional changes to how the harness files or the aria-at-app handle these assertionExceptions which is great!

lib/data/parse-test-csv-row.js
evmiguel: @howard-e addressed this in https://github.com/w3c/aria-at/pull/1053/commits/1373d76e0d2caa58c5b2f5d746e84655b90ae0d7

howard-e (approved):
Looks good! Thanks for addressing that last bit of feedback and addressing the issue with the false error being thrown. Great stuff!

howard-e:
@mcking65 this PR should now address #1046. So making it possible that instances of valid assertionIds can be listed in `*-commands.csv > assertionExceptions` without needing a priority prefix. But would instead default to the priority already noted for that assertionId from `assertions.csv` if applicable.

Could you confirm this works as expected?

PR #1075, opened 2024-05-31
Refactor Color Viewr Slider Test Plan to V2 Test Format

IsaDC:
[Preview Tests](https://deploy-preview-1075--aria-at.netlify.app)

- Convert the Color Viewer Slider test plan to v2
- Modify navigation instructions to v2 format
- Add mode switching assertion
- Add mode switching assertion to the tests.csv file
- Add mode switching assertionexception to JAWS, NVDA and VoiceOver commands


IsaDC:
@mcking65  This test plan is ready for review.

mcking65 (approved):
@IsaDC 

Thank you for the excellent work on this plan. I made some changes, and I believe it is ready to merge. You may want to give it one more read-through given the number of changes I made.

Following is a description of the 11 commits I made to the test plan.

#### Priority of aria-orientation

I reduce the priority of the orientation assertion from 'should--have' to 'may-have'. This may be a little spicy, but I think it better aligns with our definitions, especially since the slider pattern recommends supporting value changes with both horizontal and vertical arrow keys. It would be more than sufficient if screen readers only used the orientation to provide useful hints.

#### Added references

I added a row to references.csv for aria-orientation. I also added a value to the refs column for each assertion in assertions.csv.

#### Added missing commands

1. Added f and shift+f to JAWS commands
2. Added f and shift+f to NVDA commands
3. added j and shift+j to VoiceOver commands

I noticed that NVDA currently doesn't navigate to sliders with the 'f' nav key. However, based on its documentation, it seems like this is a bug.

#### Corrections to commands

In NVDA and JAWS commands, there were two Up arrow commands in the backwards navigation test instead of one. I removed the extra one.

In JAWS, there were three down arrows in the forward navigation test. We have discussed this with Vispero. In principle, we have agreement that the test should be written such that it passes when their duplicate label bug is fixed.

#### Test format adjustments

I deleted the v1 references file.

I made some changes to the presentation numbers for the first two tests in the JAWS and NVDA command files just for consistency; they didn't change the actual test rendering.

I changed how the assertion exceptions for mode switching are written to simplify the CSV files. I set the priority of the interactionModeEnabled assertion to 0 in tests.csv. That means it is not included by default in any of the command tables. Then, I added an assertion exception to the JAWS and NVDA Tab and Shift+Tab commands in reading mode so the assertion is shown only for those commands. This way of coding it eliminates the need to include exceptions in the VoiceOver commands file.

mcking65:
I made one more change ... I just remembered that VoiceOver has slider-specific commands. You can interact with a slider by pressing vo+shift+down and then you can increment or decrement with vo+arrows.

Issue #1073, opened 2024-05-22
`npm run update-reference` is out of date


howard-e:
Running `npm run update-reference <directory>` now gives the following results:

<details>

<summary>message if v1 version of pattern currently exists</summary>

```bash
$> npm run update-reference banner

> aria-at@1.0.0 update-reference
> node scripts/update-reference.js banner

Cloning the aria-practice repo.
Locating the matching example file /path/to/aria-at/scripts/tmp/examples/landmarks/banner.html.
The example html '/path/to/aria-at/scripts/tmp/examples/landmarks/banner.html' does not exist. Please enure the current example html utl is in the references.csv file.
```

</details>

<details>

<summary>message if v2 version of pattern currently exists</summary>

```bash
$> npm run update-reference alert

> aria-at@1.0.0 update-reference
> node scripts/update-reference.js alert

`example` must be defined in references.csv with the format `https://w3c.github.io/aria-practices/examples/<PATH_TO_EXAMPLE>.html`
```

</details>

[The script references the old Respec deploy for the APG site](https://github.com/w3c/aria-at/blob/9f64a43daa8b0fec371246fe66a1af9686bcdcf9/scripts/update-reference.js#L97:L99). So it has to be revised to reference examples from the new [WAI Website](https://www.w3.org/WAI/ARIA/apg/) and current `w3c/aria-practices` repo structure, and also account for the new v2 folder/files structure where applicable.

This was raised in the [May 22, 2024 CG meeting](https://www.w3.org/2024/05/22-aria-at-minutes.html#t02).

PR #1076, opened 2024-06-03
Update `update-reference.js` script to support APG site migration

howard-e:
[Preview Tests](https://deploy-preview-1076--aria-at.netlify.app)

Addresses #1073.

This PR updates the `npm run update-reference` script to support the migration for example links in references.csv now that `w3.org/WAI/ARIA/apg` is being used in place of `w3c.github.io/aria-practices`.

### DIFF
--- a/scripts/update-reference.js
+++ b/scripts/update-reference.js
@@ -1,31 +1,29 @@
-const fse = require('fs-extra');
 const path = require('path');
+const fse = require('fs-extra');
 const simpleGit = require('simple-git');
-const mustache = require('mustache');
-const np = require('node-html-parser');
 
-const tmpPath = path.join(__dirname, 'tmp');
-let repoBasePath = tmpPath;
 const cloneURL = 'https://github.com/w3c/aria-practices';
+const apgPath = path.join(__dirname, 'aria-practices');
+let apgRepositoryBasePath = apgPath;
 
 const args = require('minimist')(process.argv.slice(2), {
   alias: {
     h: 'help',
-    r: 'aria-practices-repo',
+    r: 'aria-practices-directory',
   },
 });
 
 if (args.help) {
   console.log(`
 Default use:
-  node update-reference.js exampleName [-a ariaPracticesDirectory]
-    It will copy the latest example html from the aria-practices repo into this repo.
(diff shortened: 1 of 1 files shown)


scripts/update-reference.js
```diff
@@ -86,17 +84,88 @@ async function copyExampleToRepo(exampleName) {
...
+    let exampleUrl = (
       (referencesCsv || '').split(/\r?\n/).find(s => s.startsWith('example,')) || ''
     ).split(',')[1];
+
+    // TODO: Deprecate this when all tests support the v2 format
+    // If v2 test format
+    if (exampleUrl === 'metadata')
+      exampleUrl = (
+        (referencesCsv || '').split(/\r?\n/).find(s => s.startsWith('example,')) || ''
+      ).split(',')[2];
```
stalgiag, line 96: I know this pattern existed prior to this PR but wow. It is fine for now but we would thank ourselves later if we adopted a more spacious approach to things like this, perhaps write some helper functions

stalgiag (approved):

stalgiag (approved):
Tested and works as expected. The code also makes sense. Small comment inline, not a blocker. Thank you!

scripts/update-reference.js
stalgiag: I think this repo could benefit from more utility functions in general. I see `splitLines()`, `findLineswithPrefix()`, `extractNthValue()` here for example.

howard-e: @stalgiag agreed! I had the same thoughts while doing this but was more thinking of how it will look when v1 files no longer exist. In retrospect, that will take some time and it certainly helps to create that practice from now, for this file (and others).

I pushed a change for this in https://github.com/w3c/aria-at/pull/1076/commits/b3270285438a6ae8cb3b9b083c89d9463306d95c, let me know if this aligns with your thoughts.

stalgiag: That is a major improvement. Thanks so much!

mcking65:
@ccanash  @howard-e 

Can this be merged? It is blocking work by PAC?

howard-e:
> Can this be merged? It is blocking work by PAC?

Yep, I've merged!

PR #1086, opened 2024-07-09
use a fork of cross-var with more up-to-date deps

outofambit:
[Preview Tests](https://deploy-preview-1086--aria-at.netlify.app)

While running the `yarn db-import-tests:dev` step in the local database setup instructions from a fresh clone, I was getting the following error.

```shellsession
commands.json for v2 test format may not exist
Error found: Error: ENOENT: no such file or directory, scandir '/Users/outofambit/src/meta-aria/aria-at-app/server/scripts/import-tests/tmp/build/tests'
    at Object.readdirSync (node:fs:1527:3)
    at buildTestsAndCreateTestPlanVersions (/Users/outofambit/src/meta-aria/aria-at-app/server/scripts/import-tests/index.js:118:31)
    at async importTestPlanVersions (/Users/outofambit/src/meta-aria/aria-at-app/server/scripts/import-tests/index.js:80:7)
    at async /Users/outofambit/src/meta-aria/aria-at-app/node_modules/sequelize/lib/sequelize.js:506:18
```

@howard-e and I tracked it down to being an issue with a dependency of `cross-var`, which is effectively unmaintained and uses a very old set of babel packages. We found a fork where someone had fixed those issues and that fixed my error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.0.0",
       "license": "MIT",
       "dependencies": {
-        "cross-var": "^1.1.0",
+        "cross-var": "github:radozebra/cross-var#1.1.1",
         "csv-parser": "^3.0.0",
         "csv-writer": "^1.6.0",
         "fs-extra": "^11.1.0",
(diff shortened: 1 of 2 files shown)


howard-e (approved):
LGTM also confirmed this causes no issues with the aria-at-app's imports.

A future task should follow on patching / publishing this separately for the needs of the project so this can be kept up to date.

PR #1090, opened 2024-07-18
Radio buttons for assertion verdicts

stalgiag:
[Preview Tests](https://deploy-preview-1090--aria-at.netlify.app)

This PR replaces the checkbox with "yes" & "no" radio buttons in the assertion verdict form fields. This is part of the work of addressing [aria-at-ap/#1045](https://github.com/w3c/aria-at-app/issues/1045). The related app change is [aria-at-app/#1161](https://github.com/w3c/aria-at-app/pull/1161).

Note that nested fieldset was used as it matched other UI in the form. In my testing, both screen and screenreader, this proved usable but I am open to other structures.

Now both `assertion.passed` and `assertion.result` are `Boolean | null`. The `result` can technically be a string because of the legacy `failedReason` which no current form uses. I did not remove the code associated with this. This is also why `result` has been kept instead of being removed entirely.

### DIFF
--- a/tests/resources/aria-at-harness.mjs
+++ b/tests/resources/aria-at-harness.mjs
@@ -9,7 +9,6 @@ import {
   render,
 } from './vrender.mjs';
 import {
-  AssertionResultMap,
   userCloseWindow,
   userOpenWindow,
   WhitespaceStyleMap,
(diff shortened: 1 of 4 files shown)


howard-e (approved):
Looks good, thanks! I don't see a problem with the fieldset being used.

> Now both `assertion.passed` and `assertion.result` are `Boolean | null`. The `result` can technically be a string in a because of the legacy `failedReason` which no current form uses. I did not remove the code associated with this. This is also why `result` has been kept instead of being removed entirely.

PR #1092, opened 2024-07-31
Refactor Action Menu Button Example Using aria-activedescendant Test Plan to V2 Test Format

IsaDC:
[Preview Tests](https://deploy-preview-1092--aria-at.netlify.app)


mcking65 (changes requested):
@IsaDC 

Overall, looks great. There are just a few details to address.

Since the tests include multiple elements that have names that need to be conveyed , please use the assertion format that includes the element being named. For example, instead of:

>Name 'Actions' is conveyed

I recommend the format:

>Name of the menu button, 'Actions', is conveyed

Also, please fill in the refId column in assertions.csv.

I believe JAWS and NVDA should have assertions for switching to interaction mode in:
* In the first two tests for Tab and Shift+Tab
* In Test 4: Open a menu for space and enter when in reading mode

IsaDC:
@mcking65  I have implemented the requested changes, except for adding the assertions for switching to interaction mode in Test 4: "Open a menu with space and enter when in reading mode," as these assertions already exist.
Since I am still getting familiar with the references, please feel free to correct any mistakes I may have made.
Thank you.

mcking65:
@IsaDC

Thank you for the updates! I made a few editorial changes to assertions and updated references.

I have one more suggestion for "Test 6: Request information about a menu item". 

* Assert that insert+tab in JAWS and NVDA and ctrl-opt-f3 give the name and role of the menu
* Test it in vc and browse mode as well

IsaDC:
@mcking65  I implemented the requested changes

mcking65:
@IsaDC 

I found the need for some more changes. To ensure we could start testing tomorrow, I made the changes and merged.

the changes are listed below. I don't think they are contraversial; they are all consistent with prior plans. Please review the changes. If there are problems, open a new PR.

This list could be helpful when working on the other two menu button test plan refactors.

Modified the HTML so the 'afterlink' is immediately after the menu button. Previously it was after the text field so that pressing up arrow when navigating back did not go to the menu button.

Changed one of the name assertions used in tests 4, 5, and 6 for opening a menu. I added a new priority 3 assertion 'nameMenuActions' for asserting the name of the menu.

Previously, when tests 4,5, and 6 had focus on the menu item, they used the assertion 'nameActions', which is:

>Name of the menu button, 'Actions', is conveyed

With the new assertion, those tests now have the assertion:

>Name of the menu, 'Actions', is conveyed

I changed the priority of assertions related to posinset and setsize to 2; it was set at 1.

I changed the priority of assertion 'roleMenu' to 3; it was previously 1.

In the assertions column of tests.csv, I revised the order of assertions to be consistently name, role, posinset, setsize, optional assertions, mode switch. This puts the must assertions first in every test, followed by shoulds, then by may.

Added the mode switch assertion to the assertions column in tests.csv for tests 1 and 2. It is present in the commands files for those tests, but for the first 2 tests, missing from the assertions list.

mcking65 (approved):
Thank you @IsaDC, excellent work on this. It's now ready to ship.

PR #1093, opened 2024-07-31
Assertion verdict radio buttons

stalgiag:
[Preview Tests](https://deploy-preview-1093--aria-at.netlify.app)

*Reopening after #1090 was merged prematurely*

**Leaving in `Draft`to indicate that this should be merged with caution only after the related changes are ready to be deployed on the app side. Reviewers can switch to `Ready for Review` once they have acknowledged that it is necessary to time these merges carefully. Thank you!**

[Preview Tests](https://deploy-preview-1090--aria-at.netlify.app)

This PR replaces the checkbox with "yes" & "no" radio buttons in the assertion verdict form fields. This is part of the work of addressing [aria-at-ap/#1045](https://github.com/w3c/aria-at-app/issues/1045). The related app change is [aria-at-app/#1161](https://github.com/w3c/aria-at-app/pull/1161).

Note that nested fieldset was used as it matched other UI in the form. In my testing, both screen and screenreader, this proved usable but I am open to other structures.

Now both `assertion.passed` and `assertion.result` are `Boolean | null`. The `result` can technically be a string because of the legacy `failedReason` which no current form uses. I did not remove the code associated with this. This is also why `result` has been kept instead of being removed entirely.


howard-e (approved):
Previously approved here: https://github.com/w3c/aria-at/pull/1090#pullrequestreview-2192406088

Because this change can impact the accuracy of collected results in aria-at-app, the plan is to merge this work along with https://github.com/w3c/aria-at-app/pull/1161, before a new production deployment is to be done. The current target for that is Monday August 5, 2024

PR #1095, opened 2024-08-05
Update dependencies and scripts

howard-e:
[Preview Tests](https://deploy-preview-1095--aria-at.netlify.app)

* Bumps several major dependencies
* Removes eslint-plugin-import due to current incompatibility with eslint9
* Adds additional useful scripts
* Due to dependency updates, marked engines.node as `>=20.x` which should target the latest LTS
* Supersedes current [dependabot PRs](https://github.com/w3c/aria-at/pulls/app%2Fdependabot)


stalgiag (approved):
These changes are welcome. Thank you!

PR #1096, opened 2024-08-05
Move `<root>/tests/resources` folder to `<root>/resources`

howard-e:


howard-e:
Drafting for now as this work will need a matching `w3c/aria-at-app` PR to support these changes during test plan version imports

stalgiag (approved):
This is a better location and all of the changes in this PR are straightforward in supporting that relocation. Thanks!

Issue #1089, opened 2024-07-16
Reduce duplication between v1 and v2 `process-test-directory` builders


howard-e:
[...] on duplication between `lib/data/process-test-directory.js` and `lib/data/process-test-directory-v1.js` [...]:

> v1 csv resources and references eventually won't exist so there won't be a need to build them. A couple v1 files have already been removed since introducing v2
>
> So separating that into dedicated orchestrating scripts for the v1 and v2 test formats rather than creating branches in a single file that we may want to eventually prune in the future seemed much easier to do
>
> Some parts are similar enough though that the functions could be consolidated in a single util and used separately
>
> [...]

_Originally posted by @jugglinmike in https://github.com/w3c/aria-at/pull/907#pullrequestreview-1905183382_

PR #1107, opened 2024-08-15
Reduce duplication in `process-test-directory` builders

howard-e:
[Preview Tests](https://deploy-preview-1107--aria-at.netlify.app)

Addresses #1089. Branched off of work in #1105.


stalgiag (approved):
This makes sense to me and the new utilities are great connection points. I used this [diff viewer](https://www.diffchecker.com/xik8gW3b/) to help me understand the changes and all makes sense. Thanks!

howard-e:
Thanks! Reconfirmed this doesn't causes any [build issues on aria-at-app](https://github.com/w3c/aria-at-app/compare/test-ariaat-1107) with a new import and verified locally no oddities happen when doing a local import

PR #1108, opened 2024-08-15
Update error message on "Unwanted U+FEFF" found in *.csv files

howard-e:
[Preview Tests](https://deploy-preview-1108--aria-at.netlify.app)

An error message was being thrown when running `npm run build` which read:

```
Unwanted U+FEFF found for key ﻿testId at index 0 while processing CSV.
Unwanted U+FEFF found for key ﻿assertionId at index 0 while processing CSV.
```

Found this to be coming from `tests/menu-button-actions-active-descendant/data/tests.csv` and `tests/menu-button-actions-active-descendant/data/assertions.csv`. This PR updates the message to add more detail and to remove the invisible character from both files:

```
[tests/menu-button-actions-active-descendant]: Unwanted U+FEFF found for key ﻿testId at index 0 while processing tests.csv.
[tests/menu-button-actions-active-descendant]: Unwanted U+FEFF found for key ﻿assertionId at index 0 while processing assertions.csv.
```

Seems this was introduced in #1092.

### DIFF
--- a/lib/data/process-test-directory-v1.js
+++ b/lib/data/process-test-directory-v1.js
@@ -611,7 +611,7 @@ ${rows}
 
   // intended to be an internal helper to reduce some code duplication and make logging for csv errors simpler
   async function readCSVFile(filePath, rowValidator = identity => identity) {
-    const rawCSV = await readCSV(testPlanRecord.find(filePath));
+    const rawCSV = await readCSV(testPlanRecord.find(filePath), testPlanName);
     let index = 0;
     function printError(message) {
       // line number is index+2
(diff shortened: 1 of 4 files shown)


IsaDC:
Thank you @howard-e 
That error confused me a few times.

nschonni:
May want to update https://github.com/w3c/aria-at/blob/master/.gitattributes to opt everything into `eol=lf`

howard-e:
@IsaDC said:

> Thank you @howard-e That error confused me a few times.

No problem! I also confirmed that this won't create a new version in the app because the test content hasn't actually changed so if all is well to you here, feel free to merge!

howard-e:
> May want to update https://github.com/w3c/aria-at/blob/master/.gitattributes to opt everything into `eol=lf`

@nschonni certainly could! Could expand to include *.csv as well. Although the relation between `eol` and `u+feff` is lost on me here. Is there?

nschonni:
You're right, it looks like there isn't anything native to git that will clean out the BOM. The `encoding` is set in `.editorconfig`, but I'm not sure how that would be enforced

stalgiag (approved):
Thanks for this! I needed to use the changes in this PR to make sense of the error I was seeing in the console in #1105 . I can confirm that these changes are helpful and necessary!

PR #1112, opened 2024-08-21
Add `*.csv` to .gitattributes

howard-e:
Continued discussion from:
* https://github.com/w3c/aria-at/pull/1108#issuecomment-2292336734
* https://github.com/w3c/aria-at/pull/1108#issuecomment-2296957598


nschonni:
Yeah, something like https://www.npmjs.com/package/strip-bom-cli with husky + CI doublechecking would probably work

Issue #1002, opened 2023-10-23
v2 Test Format: Space-separated settings in AT_KEY-commands.csv


stalgiag:
We have been working to merge the work in #997 and make adjustments to aria-at-app in preparation for the v2 test format landing. As we are doing that work, we have read in Howard's notes that one aspect of the v2 test format has not yet been accounted for in that PR, space-separated listing of setting names in AT_KEY-commands.csv. 

The reasoning given for the absence of this feature is two-fold. One is that it would increase the complexity involved in building the review and test files. Second, there is a sense that the use-case this feature aims to support is already supported.

Howard provided two possibilities for how this use-case might be supported without adding the space-separated settings list in AT_KEY-commands.csv.
1. Test authors could write additional rows for each settings value using the same whole number in the `presentationNumber`. I believe Howard's thinking was that these additional rows would exist in AT_KEY-commands.csv. The proposed authoring might resemble this:
```csv
testId,command,assertionExceptions,presentationNumber
reqInfoAboutBreadcrumbLink,virtualCursor,,7
reqInfoAboutBreadcrumbLink,ins+tab,,7.0
reqInfoAboutBreadcrumbLink,ins+up,,7.1
reqInfoAboutBreadcrumbLink,pcCursor,,8
reqInfoAboutBreadcrumbLink,ins+tab,,8.0
reqInfoAboutBreadcrumbLink,ins+up,,8.1
```
2. Another proposal given by Howard was that "The `assertionStatement` (in assertions.csv) can still describe an additional ‘settings’ value being used without it existing in the settings column." I am less clear on what is meant by this proposal. It is possible that we could parse and remove a `settings` key from the `assertionStatement` and use this in combination with the AT to render the relevant settings instructions where applicable.

What are your thoughts on either of these approaches being used in lieu of the space-separated settings list in the AT_KEY-commands.csv files? I read the discussion in #974 to see if there is a reason for this column that both Howard and I are missing. I didn't see anything obvious but it is possible that we both missed something.

If we are mistaken and find that the space-separated settings list isn't adequately supported elsewhere, then this feature will be added in a follow-up to #997. The goal of this issue is to get the ball rolling on this discussion so that we are well-positioned to address feedback upon Howard's return.

related to #977

mcking65:
we need it to work as specified in the V2 format specification. It can be a follow-on feature. The current use case of one setting is adequate for all of our current tests.

[...]

PR #1125, opened 2024-09-03
Support multiple settings being defined in `*-commands.csv`

howard-e:
Address #1002

### DIFF
--- a/lib/data/parse-test-csv-row.js
+++ b/lib/data/parse-test-csv-row.js
@@ -149,8 +149,9 @@ function parseTestCSVRowV2({ tests, assertions, scripts, commands }) {
     }
 
     // Also account for 'modeless' AT configurations with 'defaultMode'
-    // TODO: Account for a list of settings as described by https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#settings
-    const settingsValue = command.settings || 'defaultMode';
+    const [primarySetting, ...secondarySettings] = command.settings
+      ? command.settings.split(' ')
+      : ['defaultMode'];
 
     if (isEmptyObject(testsParsed[arrPosition])) {
       testsParsed[arrPosition] = {
@@ -162,7 +163,8 @@ function parseTestCSVRowV2({ tests, assertions, scripts, commands }) {
           ats: [
             {
               ...atTargetInfo,
-              settings: settingsValue,
+              settings: primarySetting,
+              secondarySettings,
             },
           ],
         },
@@ -178,12 +180,13 @@ function parseTestCSVRowV2({ tests, assertions, scripts, commands }) {
     } else {
       if (
         !testsParsed[arrPosition].target.ats.find(
-          e => e.key === atTargetInfo.key && e.settings === settingsValue
+          e => e.key === atTargetInfo.key && e.settings === primarySetting
         )
       )
         testsParsed[arrPosition].target.ats.push({
           ...atTargetInfo,
-          settings: settingsValue,
+          settings: primarySetting,
+          secondarySettings,
         });
     }
   }
@@ -203,25 +206,28 @@ function parseTestCSVRowV2({ tests, assertions, scripts, commands }) {
       return (
         testId === commandInfo.testId &&
         command === commandInfo.command &&
-        settings === commandInfo.settings &&
+        settings === commandInfo.settings && // the primary settings value
         presentationNumber === commandInfo.presentationNumber
       );
     }
 
     for (const command of commands) {
       testsParsed.forEach(test => {
+        const [primarySetting, ...secondarySettings] = command.settings
+          ? command.settings.split(' ')
+          : ['defaultMode'];
+
         if (
           command.testId === test.testId &&
-          test.target.ats.some(
-            at => at.key === key && at.settings === (command.settings || 'defaultMode')
-          )
+          test.target.ats.some(at => at.key === key && at.settings === primarySetting)
         ) {
           const commandInfo = {
             testId: command.testId,
             command: command.command,
-            settings: command.settings || 'defaultMode',
+            settings: primarySetting,
             presentationNumber: Number(command.presentationNumber),
             assertionExceptions: command.assertionExceptions,
+            secondarySettings,
           };
 
           // Test level commandInfo, useful for getting assertionExceptions and
--- a/tests/resources/aria-at-test-io-format.mjs
+++ b/tests/resources/aria-at-test-io-format.mjs
@@ -898,9 +898,10 @@ class BehaviorInput {
       { task: info.task || info.testId },
       target.mode || target.at.settings
     );
-    commandsAndSettings = commandsAndSettings.map(cs => {
+    commandsAndSettings = commandsAndSettings.map((cs, index) => {
       return {
         ...cs,
+        presentationNumber: Number(commands[index].presentationNumber),
         settingsInstructions: cs.settingsInstructions
           ? cs.settingsInstructions.map(el => normalizeString(el))
           : null,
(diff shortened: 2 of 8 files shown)


howard-e:
cc @mcking65

Thelucas504 (approved):

lib/data/parse-test-csv-row.js
```diff
@@ -203,25 +204,28 @@ function parseTestCSVRowV2({ tests, assertions, scripts, commands }) {
...
-          test.target.ats.some(
-            at => at.key === key && at.settings === (command.settings || 'defaultMode')
-          )
+          test.target.ats.some(at => at.key === key && at.settings === primarySetting)
         ) {
           const commandInfo = {
             testId: command.testId,
             command: command.command,
-            settings: command.settings || 'defaultMode',
+            settings: primarySetting,
```
jugglinmike, line 225: This definition of `settings` isn't really appropriate in the presence of `additionalSettings`. Will this change also be reflected in [the test format definition](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2)? What would be the ramifications of renaming this property to `primarySetting`?

howard-e: > This definition of `settings` isn't really appropriate in the presence of `additionalSettings` ... renaming this property to `primarySetting`?

I agree! I figured as much when sharing the initial [gist on what's different here](https://gist.github.com/howard-e/89a06d917946f5e922740dccee08d4e2). I proposed `mainSetting` then and have since used `primarySetting` here as well which feels more appropriate. Otherwise, it certainly presents a bit of confusion without this context.

> Will this change also be reflected in [the test format definition](https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2)?

I don't think it necessary. Seems to be something to consider as an implementation detail.

> What would be the ramifications ...

A widespread refactor across the references in this repository, aria-at-app (as well as some migration or backward compatibility work) and also in aria-at-automation-harness and related resources I assume. That will certainly need a bit of coordination that I'm uneasy trying to get that done now. Do you feel strongly otherwise?

jugglinmike: I agree that it's an implementation detail for the Community Group at large. That said, the amount of inter-project coordination needed to support change makes it feel like a public-facing design.

This change creates a gap between what the CG calls "settings" and what code maintainers call "settings", and to me, that's a problem which is distinct from the internally inconsistent names.

I dropped in from https://github.com/w3c/aria-at-automation-harness/pull/73 to verify that my understanding of the situation was accurate. I trust your judgement when it comes to coordinating changes; I'd only advocate for prioritizing that refactoring sooner rather than later!

tests/resources/aria-at-test-io-format.mjs
```diff
@@ -375,58 +375,58 @@ class CommandsInput {
     return commands;
   }
 
-  getCommandsV2({ task }, mode) {
+  getCommandsV2(task, mode) {
```
stalgiag, line 378: 👍

tests/resources/aria-at-test-io-format.mjs
```diff
@@ -932,7 +953,24 @@ class BehaviorInput {
...
+          for (const additionalSetting of foundCommandInfo.additionalSettings) {
+            const expandedSettings = {
+              settings: additionalSetting,
+              settingsText: target.at.raw.settings[additionalSetting].screenText,
+              settingsInstructions: target.at.raw.settings[additionalSetting].instructions.map(el =>
+                normalizeString(el)
+              ),
+            };
+            additionalSettingsExpanded.push(expandedSettings);
+          }
```
stalgiag, line 966: Wonder if this can be cleanly combined with the similar code above?

tests/resources/aria-at-test-run.mjs
```diff
@@ -185,10 +201,24 @@ export function instructionDocument(resultState, hooks) {
...
-              screenText && settings !== 'defaultMode' ? ` (${screenText})` : ''
-            }`;
+            const {
+              description: settings,
+              text: screenText,
+              additionalSettingsExpanded = [],
+            } = commandSettings[index];
+
+            const hasScreenText = screenText && settings !== 'defaultMode';
+            const hasadditionalSettings = hasScreenText && additionalSettingsExpanded.length > 0;
```
stalgiag, line 211: I go back and forth on this. I get the logic of keeping the camel case from the `additionalSettings` but having a variable without normal capitalization does feel odd

tests/resources/aria-at-test-run.mjs
```diff
@@ -229,13 +259,27 @@ export function instructionDocument(resultState, hooks) {
...
+    const hasScreenText = settingsText && settings !== 'defaultMode';
+    const hasadditionalSettings = hasScreenText && additionalSettingsExpanded.length > 0;
+
+    let screenTextRender = hasScreenText ? ` (${settingsText}` : '';
+    if (hasadditionalSettings) {
+      additionalSettingsExpanded.forEach(({ settingsText }) => {
+        screenTextRender = `${screenTextRender} and ${settingsText}`;
+      });
+    }
+    screenTextRender = `${screenTextRender})`;
```
stalgiag, line 279: Could also possibly be combined with the similar code above?

stalgiag (approved):
Looks great! I reviewed locally and tested making changes. The additional settings render as expected. Left a few comments inline but nothing blocking. Thanks Howard!

lib/data/parse-test-csv-row.js
```diff
@@ -149,8 +149,9 @@ function parseTestCSVRowV2({ tests, assertions, scripts, commands }) {
     }
 
     // Also account for 'modeless' AT configurations with 'defaultMode'
-    // TODO: Account for a list of settings as described by https://github.com/w3c/aria-at/wiki/Test-Format-Definition-V2#settings
-    const settingsValue = command.settings || 'defaultMode';
+    const [primarySetting, ...additionalSettings] = command.settings
+      ? command.settings.split(' ')
```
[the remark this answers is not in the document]
howard-e: Relevant to my reply [here](https://github.com/w3c/aria-at/pull/1125#discussion_r1939608816), I agree with that point. I've reverted this approach

tests/resources/aria-at-test-io-format.mjs
howard-e: This block of code has been removed so this shouldn't be relevant right now

tests/resources/aria-at-test-run.mjs
howard-e: Thanks! This has also been removed!

tests/resources/aria-at-test-run.mjs
howard-e: This has been changed slightly. Do you still feel the same here?

stalgiag: I don't feel strongly but I am looking at those lines that are identical and wondering if a `getScreenTextRender()` might replace them

```
const hasScreenText = settingsText && settings !== 'defaultMode';
    const hasadditionalSettings = hasScreenText && additionalSettingsExpanded.length > 0;

    let screenTextRender = hasScreenText ? ` (${settingsText}` : '';
    if (hasadditionalSettings) {
      additionalSettingsExpanded.forEach(({ settingsText }) => {
        screenTextRender = `${screenTextRender} and ${settingsText}`;
      });
    }
    screenTextRender = `${screenTextRender})`;
```

howard-e: Done in https://github.com/w3c/aria-at/pull/1125/commits/d3e5b1481961af508472cc7ba50bb25e49d6f8fe.

Note I: That snippet posted above was changed as well since I've dropped the "additional settings" concept but the reusable utilities is definitely the direction to go for a lot of this file.
Note: Also planning to bring to wider changes to promote more reusability and clean up through #1180.

stalgiag: Oh! My apologies. Thanks for addressing even though it was outdated haha! 😅

stalgiag (approved):
Thanks for taking the time to rework this is! I agree with the decisions made here and have confirmed this continues to work as expected.

PR #1128, opened 2024-09-23
Include type changes for additionalSettings

outofambit:
[Preview Tests](https://deploy-preview-1128--aria-at.netlify.app)

For https://github.com/w3c/aria-at/pull/1125

[Preview Tests](https://deploy-preview-1128--aria-at.netlify.app)

### DIFF
--- a/types/aria-at-file.js
+++ b/types/aria-at-file.js
@@ -97,7 +97,7 @@
  * @property {string} commands[].keypresses[].keystroke single human-readable key or key chord press
  * @property {string} [commands[].extraInstruction] human-readable additional instruction to follow
  * @property {string} [commands[].settings] this property only exists on v2 tests
- * @property {string[]} [commands[].settings] this property only exists on v2 tests. settings in addition to the single option in `setting`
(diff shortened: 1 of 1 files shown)


types/aria-at-file.js
```diff
@@ -97,6 +97,7 @@
  * @property {string} commands[].keypresses[].keystroke single human-readable key or key chord press
  * @property {string} [commands[].extraInstruction] human-readable additional instruction to follow
  * @property {string} [commands[].settings] this property only exists on v2 tests
+ * @property {string[]} [commands[].settings] this property only exists on v2 tests. settings in addition to the single option in `setting`
```
jugglinmike, line 100:
```suggestion
 * @property {string[]} [commands[].additionalSettings] this property only exists on v2 tests. settings in addition to the single option in `setting`
```

howard-e:
@outofambit following the changes I made in #1125 and also explained in https://github.com/w3c/aria-at/pull/1125#discussion_r1939608816, this PR should no longer be necessary but thanks for making this potential fix at the time, to that work!

This also means same should be reflected in https://github.com/w3c/aria-at-automation-harness/pull/73

PR #1156, opened 2024-11-06
Refactor Radio Group Example Using Roving tabindex Test Plan to V2 Test Format

IsaDC:
- Convert previous test plan version to v2 format
- Remove previous test plan csv files and log
- update assertions.csv file
- Update tests.csv file
- Update ATs commands
- Add htmlLin to references.csv
- Update references.csv


IsaDC:
@mcking65  Ready for review.

mcking65:
@IsaDC 

1. The setup script is not setting focus for tests 5 through 15. It's probably an ID issue. Focus seems to be set on the body. Note: I made sure window is maximized.
2. Test 7: NVDA is missing down arrow with browse mode on.
3. Test 8: NVDA is missing up arrow with browse mode on.
4. Test 9: NVDA is missing down arrow with browse mode on.
5. Test 10: NVDA is missing up arrow with browse mode on.
6. Test 11: JAWS should have a mode switch assertion for the enter key. Note: NVDA does not switch modes, and I assume that is an intentional NVAccess decision, so we should not assert it for NVDA. Unlike some other composite widgets, there valid arguments for either behavior with radio buttons.

mcking65 (commented):
@IsaDC 

After fixing and verifying the 6 items I mentioned in my prior comment, please go ahead and merge.

IsaDC:
h> @IsaDC
> 
> 1. The setup script is not setting focus for tests 5 through 15. It's probably an ID issue. Focus seems to be set on the body. Note: I made sure window is maximized.
> 2. Test 7: NVDA is missing down arrow with browse mode on.
> 3. Test 8: NVDA is missing up arrow with browse mode on.
> 4. Test 9: NVDA is missing down arrow with browse mode on.
> 5. Test 10: NVDA is missing up arrow with browse mode on.
> 6. Test 11: JAWS should have a mode switch assertion for the enter key. Note: NVDA does not switch modes, and I assume that is an intentional NVAccess decision, so we should not assert it for NVDA. Unlike some other composite widgets, there valid arguments for either behavior with radio buttons.

Hi @mcking65 
in the Radio Group Example Using aria-activedescendant Test Plan, up/down arrow commands for NVDA are also missing (see comments 2-5). Should we add them to both?

IsaDC:
@jscholes Please take a look at the setupScripts that set focus on radio buttons—they don't seem to be working correctly and need fixing.

jscholes:
@IsaDC Fixed.

PR #1165, opened 2024-12-03
Implement Feedback: Request Info Slider Test

IsaDC:
[Preview Tests](https://deploy-preview-1165--aria-at.netlify.app)

Resolves #1145

PR #1166, opened 2024-12-04
Ignore variable values when checking mocks with `npm test`

howard-e:
[Preview Tests](https://deploy-preview-1166--aria-at.netlify.app)

This is to avoid unnecessary snapshot failures, such as in #1165

* Ignore `lastEdited` value when creating review pages in test
* Ignore `commit` and `commitDescription` values when creating review index page in test

Those values had to be derived from the source folder rather than the tests' mock data so no need to track


IsaDC (approved):

IsaDC:
Thanks @howard-e! This looks good, and fixes the issue we discussed. Approved.

Issue #1194, opened 2025-02-05
Test Feedback: Redefine assertions for "Activate a link in a dropdown" (Disclosure Navigation Menu Example, Test 14, V24.10.10)


mcking65:
It is not clear what behaviors to expect when a same page link is activated.

Same page links can refer to any of:
* An empty element
* An element containing a text node
* An element with an easily conveyed name and role, e.g., a heading.
* An element containing multiple semantic elements

Currently, this test asserts that activating the link will cause screen readers to convey information about a heading that appears at the link target. However, the heading is not the link target. The target is a div that contains the heading and content that follows the heading.

Ultimately, users need to know:
1. The link was activated.
2. The reading cursor or focus is moved.
3. Sufficient information about the new location.

Unknown: Does focus, i.e., document.activeElement, always move to a link target?

Assertions should be based on expected behaviors that:
1. Are applicable to any same page link.
2. Are applicable to any screen reader.
3. Support the user needs.

Options:
1. A single simple assertion, e.g., "The new location is conveyed"
2. A set of assertions that require information about a specific element to be conveyed based on some consensus logic. Example: Convey first non-empty element at or immediately following the target.

## Test Setup

- Test File: [test-65-activateLinkDropdown-jaws.collected.html](https://aria-at.netlify.app/tests/disclosure-navigation/test-65-activateLinkDropdown-jaws.collected.html)
- Report Page: [Link](https://aria-at.w3.org/report/163623/targets/300#result-NGUyYeyIxMiI6NzEzfQjUwYW)
- AT: JAWS (Version 2024.2406.121)
- Browser: Chrome (Version 130)

IsaDC:
@mcking65 can we close this?

PR #1204, opened 2025-03-03
Redefine assertions for Activate a link in a `dropdown`

IsaDC:
[Preview Tests](https://deploy-preview-1204--aria-at.netlify.app)

Solves #1194


mcking65 (approved):
@IsaDC  I changed assertion priority from may to must for the two assertions in the activate link test. Those are the only two assertions. I believe they meet the criteria for "must--have" behaviors. I think this is ready to ship.

PR #1217, opened 2025-03-24
chore: use new ESLint config/ignore syntax

nschonni:
[Preview Tests](https://deploy-preview-1217--aria-at.netlify.app)

Use the newer helpers to make the Ignores easier to understand, and disable the warnings around using `fetch` in Node.js 20


howard-e (approved):
Thanks @nschonni! I went ahead and merged the latest from master and also bumped `@eslint/js`

PR #1219, opened 2025-03-24
ci: pin GitHub Actions to SHAs

nschonni:
There was a ecosystem issue by tj-actions the other week which caused secrets to be spilled from CI logs. That action didn't affect this repo, but since it is part of the recommended hardening, I run `npx pin-github-action .github/workflows/` to pin them. Dependabot should still create PRs to bump them as needed

### DIFF
--- a/.github/workflows/generate-and-commit-files.yml
+++ b/.github/workflows/generate-and-commit-files.yml
@@ -13,13 +13,13 @@ jobs:
         os: [ubuntu-latest, windows-latest]
 
     steps:
-      - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


howard-e:
@nschonni very interesting! I had only seen this notice in passing but didn't check further.

Thanks for sharing [this](https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066). As stated, this repo doesn't seem affected but will monitor the discourse around it and see if we should move this forward (or in any other repos, ha)

PR #1220, opened 2025-03-24
Run CI in verbose mode by default

howard-e:
[Preview Tests](https://deploy-preview-1220--aria-at.netlify.app)

* Run `create-all-tests` in verbose by default so the error is properly reflected.
* Fix call to scoped `LOGGER` in `readCSVFile`.


stalgiag (approved):
Looks good to me! Definitely prefer the verbose output in CI

PR #1223, opened 2025-03-26
chore: add workflow to notify app when changes have been made to tests

stalgiag:
This PR adds a workflow that notifies the app repo about any updates to the test folder. This does not have a secret currently and will be left in `DRAFT` without the secret until the approach is broadly agreed upon.

related to [aria-at-app/#1359](https://github.com/w3c/aria-at-app/pull/1359).

### DIFF
--- a/.github/workflows/notify-app-on-test-change.yml
+++ b/.github/workflows/notify-app-on-test-change.yml
@@ -2,7 +2,7 @@ name: Notify App about test changes
 on:
   push:
     branches:
-      - main
+      - master
     paths:
       - 'tests/**'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-app-on-test-change.yml
```diff
@@ -0,0 +1,37 @@
...
+        run: |
+          if git diff --name-only HEAD^ HEAD | grep -q "^tests/"; then
+            echo "changes_detected=true" >> $GITHUB_OUTPUT
+          else
+            echo "changes_detected=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Trigger aria-at-app workflow
+        if: steps.check-changes.outputs.changes_detected == 'true'
+        uses: peter-evans/repository-dispatch@v2
```
howard-e, line 28: Looks like there's a v3 this can be bumped to as well

howard-e (approved):
This looks good to me! Confirmed this approach on my own forks

PR #1237, opened 2025-04-25
Refactored Test plan for The Disclosure of Answers to Frequently Asked Questions Example

IsaDC:


mcking65:
@IsaDC 

Somehow there are 4859 files changed by this pull request. I think you might have committed a temp working directory or something like that. I can't get the files tab of this PR to function with that many files changed.

mcking65:
@IsaDC 

I pushed commits with the following changes ...

#### Changed 'DescriptionListBoundary' to "listBoundary'

Made the assertion: "List boundary is conveyed".

The description list boundary is technically more correct. However, the ARIA mapping is actually "association list" and every screen reader is handling this differently. There has been discussion in ARIA about this mapping. No one seems to be happy with it.

I am not confident it would be a good idea to fail screen readers that simply call it a list. That is what NVDA and Narrator do. JAWS still calls it a "definition list". VoiceOver calls it a description list.

Open to further discussion about this change.

#### Updated references

* Change button reference to htmlAam type.
* Added a dd reference
* Added references to assertions.

#### Editorial changes to state assertions

For consistency with prior plans and editorial guidelines, changed parentheses to single quotes so assertions of the form:

>Change in state (to expanded) is conveyed

are now:

>Change in state to 'expanded' is conveyed

IsaDC:
Thank you @mcking65 
We can discuss all of this in today's meeting.

PR #1245, opened 2025-05-19
Accordion V2

### LINKED ISSUE #423: Create tests for APG design pattern example: Accordion
#### Applies To

* APG design pattern: [Accordion (Sections With Show/Hide Functionality)](https://w3c.github.io/aria-practices/#accordion)
* Specific example: [Accordion Example](https://w3c.github.io/aria-practices/examples/accordion/accordion.html)

#### Testing Notes

This issue thread will house all documentation relating to the development of a test plan for the "accordion" APG example, which conforms to the "Accordion" design pattern.  All follow-ups (including the test plan itself) will be added as new comments to facilitate notifications, but this initial comment will be updated with relevant details and URLs as needed for convenience.

#### Additional References

* [heading (role)](https://w3c.github.io/aria/#heading)
* [region (role)](https://w3c.github.io/aria/#region)
* [aria-controls (property)](https://w3c.github.io/aria/#aria-controls)
* [aria-disabled (state)](https://w3c.github.io/aria/#aria-disabled)
* [aria-expanded (state)](https://w3c.github.io/aria/#aria-expanded)

IsaDC:
[Preview Tests](https://deploy-preview-1245--aria-at.netlify.app)

Resolves #423


mcking65:
### For May 21 Meeting Discussion

What should assertion priorities be for ...
1. Assertions about heading level
2. Assertions about the heading when tabbing to the button inside the heading
3. Assertions about region when tabbing

IsaDC:
> @IsaDC
> 
> Note that I pushed a commit with some editorial changes to the assertion statements and phrases. I described the changes in the commit statement.
> 
> Could you please add references to assertions? We should have references for HTML button, aria-expanded, aria-required, HTML input type text, and region.

Sure @mcking65 
I'll add them now.

IsaDC:
@mcking65  References and assertions ready for your review.

IsaDC:
@mcking65  Commands are ready for your review.
Note: should we add "e" and "f" for navigating into the accordion pannel?

mcking65:
@IsaDC 

Because the 'b' and 'f' commands are like 'tab' in that the reading cursor is jumping, I'm thinking quick nav keys that jump to the button should have the same assertion priorities as 'tab'.

Obviously, 'h' and '3' are different from 'b' and 'f' because they jump to the heading itself.

mcking65:
@IsaDC 

In "Test 7: Navigate into an accordion panel", I suggest adding quick nav keys that jump to the edit field. I think they would have the same assertions as tab.

I believe the commands would be ..

* 'e' and 'f' for JAWS and NVDA
* 'f' and 'j' for VoiceOver.

IsaDC:
@mcking65  I believe we covered everything.

mcking65:
@IsaDC 

I suggest that we remove Up Arrow from "Test 2: Navigate backwards to an expanded accordion header".
Up arrow would only work if we put focus on the input or put alink before the header.

However, I'd like to suggest that we remove the links for navigate forward and back from this test plan.

Instead, we could set focus on the "Personal Information" header for:
* Test 1: Navigate forwards to an expanded accordion header
* Test 3: Navigate forwards to a collapsed accordion header
And change the test so the user is navigating to the "Billing Address" header.
That is, for tests 1 and 3, just inverse  what you are doing in tests 2 and 4.

This would have the advantage of catching a VoiceOver bug the current plan can't capture now. The 'b' quick nav key gets stuck inside the first header and won't navigate to the next one.

IsaDC:
> @IsaDC
> 
> I suggest that we remove Up Arrow from "Test 2: Navigate backwards to an expanded accordion header". Up arrow would only work if we put focus on the input or put alink before the header.
>
Yes, I meant to remove that. 
> However, I'd like to suggest that we remove the links for navigate forward and back from this test plan.
>
I suggest we keep them within the example, if only for consistency.
> Instead, we could set focus on the "Personal Information" header for:
> 
> * Test 1: Navigate forwards to an expanded accordion header
> * Test 3: Navigate forwards to a collapsed accordion header
>   And change the test so the user is navigating to the "Billing Address" header.
>   That is, for tests 1 and 3, just inverse  what you are doing in tests 2 and 4.
> 
> This would have the advantage of catching a VoiceOver bug the current plan can't capture now. The 'b' quick nav key gets stuck inside the first header and won't navigate to the next one.

For that we would have to modify the current example, because the first accordion header is expanded by default. Is that OK?

mcking65:
Yes, it would be fine to have the script collapse the first header and, as needed, also  expand the second header. That is no different from the type of operations we perform in other tests, such as opening a menu.

IsaDC:
@mcking65  I've implemented the requested changes.

IsaDC:
Thank you @mcking65  this is ready for review. The setupScript works for me. Please let me know if you find any issues.

IsaDC:
> @IsaDC
> 
> I confirmed from the files tab that there are a bunch of files in the build and build/review directories included in this PR that should not be included.

Thankyou @mcking65. I believe this is an issue that the Bocoup team needs to address. My process for refactoring test plans hasn't changed, and these files aren't part of my usual workflow.

jugglinmike:
Hi there, @IsaDC. Short version: this pull request is clean, now, and we're making arrangements to prevent this problem from occurring again in the future.

Long version: during [today's Community Group meeting](https://www.w3.org/2025/05/29-aria-at-minutes.html), I expressed suspicion that this was caused by a misconfigured GitHub Action. Since then, @howard-e and I refuted that theory by identifying [a commit from your computer which introduced the files](https://github.com/w3c/aria-at/pull/1227/commits/3cce98ac1376c9b2613e4dda3ae8e9414add33a3) (this wasn't obvious because the files were introduced over a month ago, in [a different pull request](https://github.com/w3c/aria-at/pull/1227)). [We're going to add a new check which will report a failure when any pull request includes files in the "build" directory.](https://github.com/w3c/aria-at/pull/1250) This won't tell us why your system is adding those files, but it will help us keep the files out. Also, if/when the problem occurs again, we'll know right away, so we can jump in and investigate.

mcking65 (approved):
Awesome!

PR #1247, opened 2025-05-21
Fixes assertion validation errors that were causing builds to hang

### LINKED ISSUE #1244: Assertion validation causing build script to hang indefinitely
For every row in `tests.csv`, the format of the `assertions` column is validated against the following regular expression:

`^(([0123]:)?[a-zA-Z-\d]+\s*)+$

Unfortunately, certain strings cause the regexp parser to get stuck in a loop of possible matches against the pattern, meaning that for all practical purposes the script seems to hang forever.  In a recent PR where this issue was uncovered (#1237), the GitHub Actions runner was stuck on this for over three hours before I killed it.

Here's a simplified test plan sample that demonstrates the problem.  Unzip it into the tests directory and run the `create-all-tests` script over it.

[hang-sample.zip](https://github.com/user-attachments/files/19911389/hang-sample.zip)

This doesn't only effect Node.  For example, this code also results in a hang in Python:

```python
import re
pattern = r'^(([0123]:)?[a-zA-Z-\d]+\s*)+$'
text = 'firstAssertion secondAssertion thirdAssertion? fourthAssertion'
re.match(pattern, text)
```

#### Potential solution
I'm not convinced that validating the entire column text upfront is that necessary.  Presumably the individual parts, as separated by whitespace, could be validated later.

CC @howard-e

ChrisC:
[Preview Tests](https://deploy-preview-1247--aria-at.netlify.app)

Allows usage of `:` characters in assertion columns. This also lightly refactors assertion validation (mostly just moves scope of functions) for easier & isolated testability. 

Ideally, we would have more test cases for all of the validators, but I'm limiting this PR to known issues for now. 

closes #1244


howard-e:
> @howard-e see my note about possibly updating `.prettierignore` on [c4504b5](https://github.com/w3c/aria-at/commit/c4504b5eea5fd0ea303bb21fb6728d52c91defe5)

Just noticed as well. Opened #1253 to track

ChrisC:
@howard-e Not sure if you've had to time to look at this yet but I think this is ok to merge?

howard-e (commented):
Looks good @ChrisC! The re-organization was well needed.

I've only left a comment below on a case where there could be an unreported build failure, which I think should be handled.

lib/data/process-test-directory/index.js
```diff
@@ -30,6 +30,116 @@ const {
...
+// Regex Formats - Common regex patterns found in *.csvs
+const numericKeyFormat = /^_(\d+)$/;
+const idFormat = /^[A-Za-z0-9-?]+$/;
+const validReferenceKeys = /^(?:refId|type|value|linkText)$/;
+const validCommandKeys = /^(?:testId|command|settings|assertionExceptions|presentationNumber)$/;
+const assertionsExceptionsFormat = /^([0123]:[a-zA-Z-\d]+\s*)+$/;
+const settingsFormat = /^[A-Za-z0-9-\s]+$/;
+const validTestsKeys = /^(?:testId|title|presentationNumber|setupScript|instructions|assertions)$/;
+const titleFormat = /^[A-Z]([A-Za-z-',\s]){2,}[^.]$/;
+const assertionsFormat = /^([A-Za-z0-9-:]+\s*)+$/;
```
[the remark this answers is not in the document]
ChrisC: Actually looks like limiting the numeric chars breaks the build. See the failure here: https://github.com/w3c/aria-at/actions/runs/15583198394/job/43883112599?pr=1247#step:5:305

```
WARNING: Error parsing /home/runner/work/aria-at/aria-at/tests/horizontal-slider/data/tests.csv line 2: Error: assertions does not match the expectedFormat: roleSlider nameRed value128 orientationHorizontal minimumValue0 maximumValue255 0:interactionModeEnabled
WARNING: Error parsing /home/runner/work/aria-at/aria-at/tests/horizontal-slider/data/assertions.csv line 2: Error: assertionId does not match the expected format: maximumValue255
```

There's an assertion ID named `maximumValue255`, so I think we don't actually want to limit numeric chars. Let me know if you disagree?

lib/data/process-test-directory/index.js
```diff
@@ -30,6 +30,116 @@ const {
...
+  if (!idFormat.test(row.testId))
+    throw new Error('testId does not match the expected format: ' + row.testId);
+  if (!titleFormat.test(row.title))
+    throw new Error('title does not match the expected format: ' + row.title);
+  if (row.assertions && !assertionsFormat.test(row.assertions))
+    throw new Error('assertions does not match the expectedFormat: ' + row.assertions);
+  if (!Number(row.presentationNumber) > 0)
+    throw new Error(
+      'presentationNumber does not match the expected format: ' + row.presentationNumber
+    );
```
howard-e, line 94: @ChrisC replying to https://github.com/w3c/aria-at/pull/1247#discussion_r2139866711:

> There's an assertion ID named maximumValue255, so I think we don't actually want to limit numeric chars. @howard-e Let me know if you disagree?

Yep, we definitely don't want that since the regex covers the entire value, so agreed that changes in the last commit should be reverted.

I was more thinking of validating a unique case like that through the `validateTestsKeys` function similar to below:

```suggestion
    );
    
    if (row.assertions) {
      const assertionIds = row.assertions.split(' ');
      for (const assertionId of assertionIds) {
        const prefixMatch = assertionId.match(/^(\d+):/);
        if (prefixMatch) {
          const priority = parseInt(prefixMatch[1]);
          if (![0, 1, 2, 3].includes(priority)) {
            throw new Error(`assertions contains invalid priority "${priority}" for assertionId "${assertionId}". Only 0, 1, 2, or 3 are allowed: ` + row.assertions);
          }
        }
      }
    }
```

What do you think?

ChrisC: This does seem little bit like over-engineering imo... besides, not restricting the numeric chars gives testers flexibility with the priority levels should they ever need it... but if you feel strongly about this, I can commit this and update the tests!

lib/data/process-test-directory/index.js
```diff
@@ -86,7 +87,10 @@ function validateTestsKeys(row) {
     throw new Error('testId does not match the expected format: ' + row.testId);
   if (!titleFormat.test(row.title))
     throw new Error('title does not match the expected format: ' + row.title);
-  if (row.assertions && !assertionsFormat.test(row.assertions))
+  if (
+    row.assertions &&
+    (!assertionsFormat.test(row.assertions) || invalidPriorityFormat.test(row.assertions))
+  )
```
ChrisC, line 93: @howard-e I was able to bring back the priority validation by simply running two regex tests :-)

__test__/processTestDirectory.test.js
```diff
@@ -24,12 +24,26 @@ describe('processTestDirectory', () => {
       expect(() => validateTestsKeys(invalidRow)).toThrow();
     });
 
+    const validPriority = { ...validRow, assertions: '0:abc 1:abc 2:abc 3:abc' };
+    it('returns row on valid priority', () => {
+      expect(validateTestsKeys(validPriority)).toMatchObject(validPriority);
+    });
+
+    ['0:abc 4:abc', '3:', '4:abc', '5:abc', '6:abc', '7:abc', '8:abc', '9:abc'].forEach(
+      assertion => {
```
ChrisC, line 33: Just to make sure, these taste cases should cover the possible invalid priority formats almost exhaustively. Let me know if you can think of any other possible invalid cases we should also check for @howard-e !

lib/data/process-test-directory/index.js
ChrisC: @howard-e this should now be resolved with https://github.com/w3c/aria-at/pull/1247/commits/594d723026ea72bd78d5af095ac9bb845b430f14

lib/data/process-test-directory/index.js
howard-e: Nice!

PR #1255, opened 2025-06-06
Test Plan for the Tabs with Automatic Activation Example

### LINKED ISSUE #454: Create tests for APG design pattern example: Alert Dialog
#### Applies To

* APG design pattern: [Alert Dialog](https://w3c.github.io/aria-practices/#alertdialog)
* Specific example: [Alert Dialog Example](https://w3c.github.io/aria-practices/examples/dialog-modal/alertdialog.html)

#### Testing Notes

This issue thread will house all documentation relating to the development of a test plan for the "Alert Dialog" APG example, which conforms to the "Alert Dialog" design pattern.  All follow-ups (including the test plan itself) will be added as new comments to facilitate notifications, but this initial comment will be updated with relevant details and URLs as needed for convenience.

#### Additional References

* [alertdialog (role)](https://w3c.github.io/aria/#alertdialog)
* [alert (role)](https://w3c.github.io/aria/#alert)
* [aria-modal (property)](https://w3c.github.io/aria/#aria-modal)
* [aria-describedby (property)](https://w3c.github.io/aria/#aria-describedby)

IsaDC:
[Preview Tests](https://deploy-preview-1255--aria-at.netlify.app)

[Preview Tabs with Automatic Activation Test Plan | For Pattern: tabs-automatic-activation](https://deploy-preview-1255--aria-at.netlify.app/review/tabs-automatic-activation)

[Preview all Tests](https://deploy-preview-1255--aria-at.netlify.app)

Resolves #454


IsaDC:
@mcking65 I'm finishing up most of the changes locally. I expect to push them tomorrow morning.

So far, I've drafted the additional tests we discussed in our meeting, added the missing commands to tests 1 and 2, made some editorial changes to the assertions, and created and added the assertion for the tab panel boundary to the appropriate test.

The next steps are to:
* Write `setupScripts` for the new tests.
* Change the wording in Tests 4 and 5 from "Navigating to the next/previous tab" to "Selecting the next/prev tab."
* Create tests for navigating to the next/previous tab in virtual cursor/browse mode with the arrow keys.
* Create a test to activate a tab using the Enter and Space keys.

Do you think we should also add the "Request information about a not selected tab" test?

mcking65:
@IsaDC 

I pushed a commit to fix some problems with references.csv and the reference column in assertion.csv.

Some special characters are not rendered correctly, example: convey name of the focused tab, 'Peter MÃƒÂ¼ller'

Assertsions for names of tabs need to be separate from assertions for names of the tabpanel elements because the naming techniques are different, so the references need to be different. Tabs need to reference the HTML AAM mapping for button. The tabpanel elements are named with aria-labelledby. 

This assertion should have priority 2: convey position of the tab, 1

In Test 2: Navigate backwards into a tab list where a tab is not selected,
I think the VO command VO+Left should probably only be pressed 2 times, not 3 ... however, I didn't verify this.


In Test 6: Request information about a not selected tab in a tab list, the setup script is not setting focus on the correct tab.

In Test 7: Activate the next tab in a tab list, remove the JAWS command: 
Down Arrow (virtual cursor active)

In Test 8: Activate the previous tab in a tab list, remove the JAWS command:
Up Arrow (virtual cursor active)


In Test 12: Navigate to the previous tab in a tab list, replace the assertion:
convey state of the tab, 'not selected'
with the assertion:
convey state of the tab, 'selected'

mcking65:
@IsaDC 

I found several more issues that I overlooked in my first review and some new ones. I went ahead  and pushed commits for them. Please review my commit statements.

Here are a few I'd like to highlight.

Virst, in the menu button test plans, we gained consensus between Vispero and Apple that that it is optional to convey the name and role of the menu when focus moves to an item inside the menu. We have the same type of condition when tabbing or using a quick nav key to move focus to a tab in the tablist. So, I changed the default priority of the following assertions to 3:

>Name of the tab list, 'Danish Composers', is conveyed
>convey role 'tab list'

However, tablist has one key difference from a menu: you can browse across the tablist boundary with arrow keys. So, I pushed additional commits for arrow key navigation commands  that overrides the default priority for these assertions with a value of 2.

Note that the tablist name  assertion was missing from the tests for requesting information; I pushed another commit to add it in, but not for insert+up, of course.

Finally, I spent considerable time with VoiceOver in Safari and Chrome. There are definitely some strange bugs when setting focus on the tab panel container. When the tabpanel receives focus, frequently the VO cursor seems to be placed at the end of the tabpanel instead of the beginning. That is why it appears to require 3 instead of 2 vo-left commands. However, this is a bug. I believe the test should expect only 2 vo-left commands to be required.. I pushed a commit that changes the VO commands for navigating backward into the tablist to use only 2 vo-left commands.

IsaDC:
Thank you @mcking65!
I'm going to merge the PR now.

PR #1264, opened 2025-06-30
Improve form UI for untestable assertions

jugglinmike:
[Preview Tests](https://deploy-preview-1264--aria-at.netlify.app)

This patch is in service of https://github.com/w3c/aria-at-app/issues/1430

### DIFF
--- a/tests/resources/aria-at-test-run.mjs
+++ b/tests/resources/aria-at-test-run.mjs
@@ -255,7 +255,7 @@ export function instructionDocument(resultState, hooks) {
       },
       untestable: {
         description: [
-          `The response to '${command}' created conditions that made assertions untestable, e.g., focused the wrong element.`,
+          `Executing '${command}' affected behavior that made assertions untestable. If checked, then at least one severe negative side effect must be recorded below.`,
           {
             highlightRequired: resultStateCommand.untestable.highlightRequired,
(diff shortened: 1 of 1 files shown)


tests/resources/aria-at-test-run.mjs
```diff
@@ -377,7 +377,12 @@ export function instructionDocument(resultState, hooks) {
                   ]),
                   enabled: behavior.checked,
                   value: behavior.more.value,
-                  focus:
+                  focusImpact:
```
howard-e, line 380: The typedef for [`InstructionDocumentAssertionChoiceOptionsOptionsMore`](https://github.com/w3c/aria-at/blob/aa944ce836bffae3f45bbfa7005a82b352c3518c/tests/resources/aria-at-test-run.mjs#L1141) should be updated to include `focusImpact` and `focusDetails`

howard-e (approved):
Looks good! I was able to confirm the focus management here works as expected when testing through the aria-at-app

tests/resources/aria-at-test-run.mjs
jugglinmike: Good catch! Done!

howard-e: Thanks!

PR #1275, opened 2025-08-08
Refactor Switch Example Test Plan to V2

IsaDC:
[Preview Tests](https://deploy-preview-1275--aria-at.netlify.app)

[Preview Switch Example Test Plan](https://deploy-preview-1275--aria-at.netlify.app/review/switch)

[Preview All Tests](https://deploy-preview-1275--aria-at.netlify.app)


IsaDC:
> Option B: Require consistent treatment as a form element and as a button element?

Yes @mcking65,  I agree with you. I'll make the necessary changes to the commands for all 3 ATs.

mcking65:
@IsaDC 

This is looking great! I have only one minor suggestion.

I recommend removing the 'span' HTML AAM reference. Since the 'span' has the switch role on it, we are not testing the semantics of 'span'. The 'generic' semantic of 'span' is overridden by the 'switch' semantic.

PR #1299, opened 2025-09-17
support.json: Add HTML attribute URL fragments

mcking65:
* Adds 162 key/value pairs to the fragments module so HTML attribute mappings can be referenced.
* Corrects the key for the HTML autonomous custom elements mapping to remove spaces and change it to camel case to be consistent with other keys.

Note that the key for each attribute starts with an '@' character, e.g., "@abbr" is the key for the "abbr" attribute. I did this for two reasons:
1. Enables all keys to be sorted alphabetically and keep all the attributes separate from the elements.
2. Distinguishes attributes from elements that have the same name, e.g., "abbr".

I don't know if using the '@' symbol for this purpose is a good approach to achieving the above two objectives. I'm open to changing it if this approach is somehow problematic.


jscholes (approved):

jscholes:
@mcking65 Looks good to me.

jscholes:
@mcking65 FYI: I caught a couple of issues after posting my previous goahead comment, so have deleted it.  Just in case you see it in your email.  Issues detailed in my change request.

mcking65:
addresse feedback from James; merging.

PR #1305, opened 2025-10-03
Test Plan for the Checkbox Example (Two State) V2

IsaDC:


mcking65:
@IsaDC 

* Each name assertion should state the element to which it applies, e.g., "name of the checkbox", "Name of the group", etc. Some statements do this, but not all. Several of the phrases are missing this.
* To be consistent with radio group, the group role and name assertions should be priority 2 when navigating with quick nav and tab. Probably default them to p2 and then override for arrow keys.
* There are redundant assertions for group role/name and boundary. I suggest keeping the role/name assertions and not asserting group boundary.

mcking65:
@IsaDC, thank you for the updates. I found a few more issues. 

Please change the assertions:

>convey state 'not checked'
>convey state 'checked'

to:

>convey state of the checkbox, 'not checked'
>convey state of the checkbox,' checked'

In tests 1 through 4, VoiceOver is missing the 'x' and 'shift+x' commands.

In tests 1 and 3, when navigating with arrows, shouldn't the priority be '1' for the assertion "convey role 'group'"?

In tests 2 and 4, when navigating backward, the cursor is not traversing the group or list boundaries. So, shouldn't we remove the following assertions?
* convey role 'group'
* convey name of the group, 'Sandwich Condiments'

In tests 5 and 6, I wonder if we should have P2 assertions for the insert+tab and VO commands for the assertions:
* convey role 'group'
* convey name of the group, 'Sandwich Condiments'

There is a strong argument for these two assertions for radios because they are normally in a group. Asserting it here would imply that the screen reader should be looking for grouping of all types of inputs. It seems to be a valid requirement to me, but may be debatable.

In tests 7 and 8, remove 'Enter' from the commands. Enter is not supported for toggling a checkbox.

IsaDC:
@mcking65 ,

> In tests 1 through 4, VoiceOver is missing the 'x' and 'shift+x' commands.>

The X and shift+x commands in VoiceOver move through list items, that's why I didn't include them in the original test plan. It just so happens that they would work because the checkboxes are inside a list.

I would argue then that we should also add i and shift+i for JAWS and NVDA, except we are not testing lists.

What do you suggest?

mcking65 (approved):
@IsaDC  looks great, let's ship it!

PR #1323, opened 2025-10-27
Disclosure of Frequently Asked Questions: Correct inaccurate ARIA/HTML feature references in assertions.csv

mcking65:
[Preview Tests](https://deploy-preview-1323--aria-at.netlify.app)

* Remove 'button' reference from assertions related to aria-expanded.


mcking65:
@IsaDC @jscholes 

Test 9, 'Navigate from an expanded disclosure button to the text of a question answer' has a reference problem. The assertion references the 'p' element, but we made a mistake there. The answer is in a 'div', not a 'p'.

As a result, we don't have a consistent way of approaching this test across screen readers. The "next paragraph' approach to get consistent reading of the entire answer does not work. With JAWS and NVDA, we could determine the right number of down arrow commands, but that is not robust. The number of down arrows is arbitrary, not based on a defined line length spec.

I've been giving this some thought and have another solution to this problem -- delete test 9. I am not convinced it is absolutely necessary to test this aspect of the pattern. What do you think?

PR #1337, opened 2025-11-04
support.json: Revise HTML AAM spec ref link text

mcking65:
[Preview Tests](https://deploy-preview-1337--aria-at.netlify.app)

The link text for HTML element mappings is currently like 'button accessibility API mapping'. It does not clearly state that the element is an HTML button. This changes it to be like 'button HTML-AAM specification'. That is the name of the spec that is being referenced. This makes the link text shorter, more accurate, and easier to understand.

PR #1338, opened 2025-11-05
Support dynamic renames of `support.json/references` properties

howard-e:
[Preview Tests](https://deploy-preview-1338--aria-at.netlify.app)

#1337 change isn't being reflected when importing it into aria-at-app.

The `linkText` and `value` variables from the references for v2 tests were being transformed here and then imported on aria-at-app. That isn't expected so this PR removes that mutation happening here, and is instead handled on the review page and where applicable in the aria-at-app instead.

https://github.com/w3c/aria-at-app/pull/1617 should be merged along with this.


stalgiag (approved):
This makes sense to me. Thanks for handling this delicate process!

